PR #7, opened 2018-05-29
ModuleTree and SLangCheck interfaces

marcelo-sousa-sonarsource:


slang-api/src/main/java/com/sonarsource/slang/api/ModuleTree.java
```diff
@@ -0,0 +1,10 @@
+package com.sonarsource.slang.api;
+
+import java.util.List;
+
+/**
+ * SLang module unit.
+ */
+public interface ModuleTree extends Tree {
```
pynicolas: I don't think we need that for the moment.

slang-api/src/main/java/com/sonarsource/slang/api/ModuleTree.java
```diff
@@ -0,0 +1,10 @@
...
+  List<Tree> types();
```
pynicolas: That seems to indicate that we don't support statements at the top level.

slang-api/src/main/java/com/sonarsource/slang/api/SLangFileScannerContext.java
```diff
@@ -0,0 +1,65 @@
+package com.sonarsource.slang.api;
+
+import java.io.File;
+import java.util.Objects;
+
+/**
+ * Context injected in check classes and used to report issues.
+ */
+public interface SLangFileScannerContext {
```
pynicolas: That should be part of #3.

marcelo-sousa-sonarsource:
These changes have already been incorporated in other pull requests.

PR #17, opened 2018-06-01
Conditionals in the grammar

marcelo-sousa-sonarsource:


slang-antlr/src/main/antlr4/com/sonarsource/slang/parser/SLang.g4
```diff
@@ -127,6 +129,21 @@ expressionName
   :  identifier
   ;
 
+conditional
+  :  IF LPAREN statementOrExpression RPAREN controlBlock (ELSE controlBlock)?
+  |  MATCH LPAREN statementOrExpression RPAREN LCURLY matchCase* RCURLY
```
pynicolas: I think we should have 2 separate grammar rules for "if" and "match": it would make it slightly easier to convert the AST.

slang-antlr/src/main/antlr4/com/sonarsource/slang/parser/SLang.g4
```diff
@@ -135,6 +152,10 @@ nativeBlock
   :  LBRACK (statementOrExpression semi)* (statementOrExpression semi?)? RBRACK 
   ;
 
+terminator
```
pynicolas: Why not call it "return"?

slang-antlr/src/main/antlr4/com/sonarsource/slang/parser/SLang.g4
```diff
@@ -135,6 +160,10 @@ nativeBlock
   :  LBRACK (statementOrExpression semi)* (statementOrExpression semi?)? RBRACK 
   ;
 
+returnExpression
```
marcelo-sousa-sonarsource: This rule cannot simply be called 'return' because in ANTLR the 'symbol return conflicts with generated code in target language or runtime'

PR #24, opened 2018-06-04
 Rule: EmptyBlock

marcelo-sousa-sonarsource:


slang-checks/src/main/java/com/sonarsource/slang/checks/EmptyBlockCheck.java
```diff
@@ -0,0 +1,53 @@
...
+public class EmptyBlockCheck implements SlangCheck {
+
+  private static final String MESSAGE = "Either remove or fill this block of code.";
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(BlockTree.class, (ctx, blockTree) -> {
+      if (!(ctx.parent() instanceof FunctionDeclarationTreeImpl)) {
+        boolean noStatementsOrExpressions = blockTree.statementOrExpressions().isEmpty();
+        boolean noComments = blockTree.metaData().commentsInside().isEmpty();
```
pynicolas: We can factorize that code in a method ("hasNoComment") and call it only when `noStatementsOrExpressions` is true.

marcelo-sousa-sonarsource: Fixed at c8d367e

PR #26, opened 2018-06-05
SQ Kotlin plugin

pynicolas:


slang-kotlin/src/main/java/com/sonarsource/slang/kotlin/KotlinRulesDefinition.java
```diff
@@ -0,0 +1,40 @@
...
+package com.sonarsource.slang.kotlin;
+
+import com.sonarsource.slang.checks.CommonCheckList;
+import org.sonar.api.server.rule.RulesDefinition;
+import org.sonarsource.analyzer.commons.RuleMetadataLoader;
+
+public class KotlinRulesDefinition implements RulesDefinition {
+
+  private static final String RESOURCE_FOLDER = "org/sonar/l10n/kotlin/rules/kotlin";
+
```
christophe-zurn-sonarsource: Did you mean to put this in a 'org/sonar/l10n/slang/rules/kotlin' folder instead? I don't see the need to have twice the 'kotlin' in the path

pynicolas: It used to be a SonaQube requirement a very long time ago: one "kotlin" represents the plugin key and other one is for the language key. I think we should keep it only if we insist on being very consistent with other plugins. As this is not a requirement anymore and as this path looks really strange, I would rather change it to something simpler.

slang-kotlin/pom.xml
```diff
@@ -53,5 +63,42 @@
...
+          <pluginName>SonarKotlin</pluginName>
+          <pluginClass>com.sonarsource.slang.kotlin.KotlinPlugin</pluginClass>
+          <!--<skipDependenciesPackaging>true</skipDependenciesPackaging>-->
+          <sonarLintSupported>true</sonarLintSupported>
+          <addMavenDescriptor>false</addMavenDescriptor>
+          <sonarQubeMinVersion>6.7</sonarQubeMinVersion>
+        </configuration>
+      </plugin>
+      <!--
+      <plugin>
```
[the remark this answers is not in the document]
pynicolas: No we don't. We can remove that.

slang-kotlin/src/main/java/com/sonarsource/slang/kotlin/KotlinRulesDefinition.java
christophe-zurn-sonarsource: Oh ok. Thx for info.
I'd be fine with a simpler path if it's not required anymore :)
Let's discuss this on thursday

Issue #33, opened 2018-06-06
Syntactic equivalence on string literals


marcelo-sousa-sonarsource:
We identify the following conditions related to string literals as equivalent:

```
if (lastTokenIsSlash) {
    current.append(nextTok)
    lastTokenIsSlash = false
} else if ("\\" == nextTok)
    lastTokenIsSlash = true
else if ("\'" == nextTok) {
    state = inQuote
} else if ("\"" == nextTok) {
    state = inDoubleQuote
} else if (" " == nextTok) {
    if (lastTokenHasBeenQuoted || current.isNotEmpty()) {
        result.add(current.toString())
        current.setLength(0)
    }
} else {
    current.append(nextTok)
}
```

PR #35, opened 2018-06-06
Syntactic equivalence on string literals (#33)

christophe-zurn-sonarsource:


slang-kotlin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -175,13 +173,8 @@ private Tree createElement(@Nullable PsiElement element) {
...
     } else if (element instanceof KtStringTemplateExpression) {
-      KtStringTemplateEntry[] entries = ((KtStringTemplateExpression) element).getEntries();
-      if (entries.length == 1 && entries[0] instanceof KtLiteralStringTemplateEntry) {
-        // Non-template strings, ie. not in the form "string ${1 + 1}"
-        return new LiteralTreeImpl(metaData, '\"' + entries[0].getText() + '\"');
-      } else {
-        return new NativeTreeImpl(metaData, new KotlinNativeKind(element), list(Arrays.stream(element.getChildren())));
-      }
+      // Do not differentiate between template and non-template strings for now
+      return new LiteralTreeImpl(metaData, element.getText());
```
pynicolas: I believe that we lose some information when doing that. Ideally, when a template refers to an identifier, we should have this identifier somewhere in the AST (and it would be good to have a unit test for that). That's not a big issue for the rules which we have for the moment, so we could decide to leave that aside.

slang-kotlin/src/test/java/com/sonarsource/slang/kotlin/KotlinParserTest.java
```diff
@@ -76,6 +80,20 @@ public void testLiterals() {
       .isEquivalentTo(slangStatements("554; true; false; null; \"string\"; 'c';"));
   }
 
+  @Test
+  public void testEscapedStringLiterals() throws IOException {
+    String content = new String(Files.readAllBytes(Paths.get("src/test/resources/strings.kt")), UTF_8);
```
pynicolas: I think it would be better to not have the literals in an external file. We can maybe create a variable with an explicit name to make it understandable, e.g.: `Tree escapedBackslash = kotlin("\"\\\\\"")`.

slang-kotlin/src/test/java/com/sonarsource/slang/kotlin/KotlinParserTest.java
```diff
@@ -76,6 +77,43 @@ public void testLiterals() {
...
   }
 
+  @Test
+  public void testSimpleStringLiterals() {
+    String escapedBackslash = createEscapedString('\\');
+    String escapedSingleQuote = createEscapedString('\'');
+    String escapedDoubleQuote = createEscapedString('\"');
+    String emptyString = createString("");
+    List<Tree> literals = kotlinStatements(escapedBackslash + ";" + escapedSingleQuote + ";" + escapedDoubleQuote + ";" + emptyString + ";");
+    Tree escapedBackslashLiteral = literals.get(0);
```
pynicolas: Why not `Tree escapedBackslashLiteral = kotlinStatement(escapedBackslash);`?

slang-kotlin/src/test/java/com/sonarsource/slang/kotlin/KotlinParserTest.java
```diff
@@ -76,6 +77,43 @@ public void testLiterals() {
...
+    List<Tree> complexStrings = kotlinStatements("\"identifier ${x}\"; \"block ${1 == 1}\"");
+    Tree stringWithIdentifier = complexStrings.get(0);
+    Tree identifierExpressionContainer = stringWithIdentifier.children().get(1);
+    assertTree(stringWithIdentifier).isInstanceOf(NativeTree.class);
+    assertTree(stringWithIdentifier.children().get(0)).isLiteral("identifier ");
+    assertTree(identifierExpressionContainer).isInstanceOf(NativeTree.class);
+    assertThat(identifierExpressionContainer.children()).hasSize(1);
+    assertTree(identifierExpressionContainer.children().get(0)).isIdentifier("x");
+
+    Tree stringWithBlock = complexStrings.get(1);
```
pynicolas: Why not `Tree escapedBackslashLiteral = kotlinStatement("\"block ${1 == 1}\"");`

slang-kotlin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -187,13 +189,20 @@ private Tree createElement(@Nullable PsiElement element) {
...
-        return new NativeTreeImpl(metaData, new KotlinNativeKind(element), list(Arrays.stream(element.getChildren())));
+        String fullString = Arrays.stream(((KtStringTemplateExpression) element).getEntries())
+          .map(KtElement::getText).collect(Collectors.joining("", "\"", "\""));
+        return new LiteralTreeImpl(metaData, fullString);
       }
+    } else if (element instanceof KtLiteralStringTemplateEntry || element instanceof KtEscapeStringTemplateEntry) {
+      return new LiteralTreeImpl(metaData, element.getText());
+    } else if (element instanceof KtStringTemplateEntryWithExpression) {
+      // Covers both KtBlockStringTemplateEntry and KtSimpleNameStringTemplateEntry
+      return new NativeTreeImpl(metaData, new KotlinNativeKind(element), list(Arrays.stream(element.getChildren())));
```
pynicolas: SonarLint would tell you that this branch is the same as the one 2 lines below.

slang-kotlin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -187,13 +189,20 @@ private Tree createElement(@Nullable PsiElement element) {
...
       return new MatchCaseTreeImpl(getTreeMetaData(whenElement), conditions, body);
     } else if (element instanceof KtStringTemplateExpression) {
-      KtStringTemplateEntry[] entries = ((KtStringTemplateExpression) element).getEntries();
-      if (entries.length == 1 && entries[0] instanceof KtLiteralStringTemplateEntry) {
-        // Non-template strings, ie. not in the form "string ${1 + 1}"
-        return new LiteralTreeImpl(metaData, '\"' + entries[0].getText() + '\"');
+      boolean complexString = Arrays.stream(((KtStringTemplateExpression) element).getEntries())
+        .anyMatch(KtStringTemplateEntryWithExpression.class::isInstance);
+      if (complexString) {
+        return new NativeTreeImpl(metaData, new KotlinNativeKind(element), list(Arrays.stream(((KtStringTemplateExpression) element).getEntries())));
```
pynicolas: Maybe we can factorize a bit by creating a variable for `((KtStringTemplateExpression) element).getEntries()`. In fact, replacing this expression with `element.getChildren()` doesn't seem to break the tests...

christophe-zurn-sonarsource: True. Actually I couldn't find any case where `element.getChildren()` would result in a different Tree than using `getEntries()`. And if such a case happens, it makes sense to use the children anyway in order to have different `NativeTree` in the end

slang-kotlin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -187,13 +189,20 @@ private Tree createElement(@Nullable PsiElement element) {
...
       } else {
-        return new NativeTreeImpl(metaData, new KotlinNativeKind(element), list(Arrays.stream(element.getChildren())));
+        String fullString = Arrays.stream(((KtStringTemplateExpression) element).getEntries())
+          .map(KtElement::getText).collect(Collectors.joining("", "\"", "\""));
+        return new LiteralTreeImpl(metaData, fullString);
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: Yes. It actually result in the same thing... So I simplified to use `element.getText()`

Issue #28, opened 2018-06-05
Add support for assignments in SLang AST


marcelo-sousa-sonarsource:

pynicolas:
Fixed by #36 .

PR #36, opened 2018-06-06
Add support for assignments in SLang AST (#28)

christophe-zurn-sonarsource:


christophe-zurn-sonarsource:
Please do not squash this PR into master but rebase it instead. Each commit is an independent task, but the later ones depend on the previous ones.

PR #38, opened 2018-06-06
Support analysis of mature Kotlin projects

marcelo-sousa-sonarsource:


slang-kotlin/src/main/java/com/sonarsource/slang/kotlin/KotlinSensor.java
```diff
@@ -63,7 +63,14 @@ public void execute(SensorContext sensorContext) {
 
   private void analyseFiles(SensorContext sensorContext, Iterable<InputFile> inputFiles, List<TreeVisitor<InputFileContext>> visitors) {
     for (InputFile inputFile : inputFiles) {
-      analyseFile(sensorContext, inputFile, visitors);
+      try {
+        analyseFile(sensorContext, inputFile, visitors);
+      } catch (IllegalStateException e) {
+        // FIXME TBD
+      } catch (Throwable e) {
+        // FIXME TBD
```
christophe-zurn-sonarsource: Maybe just a comment on why we did it like this for now (assert errors following kotlin lexer exception), so that we do not forget.

slang-kotlin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -134,6 +140,9 @@ private Tree createElement(@Nullable PsiElement element) {
         // FIXME are we sure we want body of function as block tree ?
         bodyTree = new BlockTreeImpl(bodyTree.metaData(), Collections.singletonList(bodyTree));
       }
+      if (bodyTree != null && bodyTree.children().isEmpty()) {
+        bodyTree = null;
```
christophe-zurn-sonarsource: I think this if can be integrated somewhat with the previous one.
Also, a comment on why we do it would be nice too. I know this is related to empty lambda, but it might not be obvious to someone who did not investigate the issue.

marcelo-sousa-sonarsource:
I have addressed all the comments in d6b4805.

pynicolas:
Each change in the "production" code (src/main) should have a corresponding change in the automated tests (src/test).

PR #42, opened 2018-06-07
Rule: All branches in a conditional structure should not have exactly…

pynicolas:
… the same implementation


slang-checks/src/main/java/com/sonarsource/slang/checks/AllBranchesIdenticalCheck.java
```diff
@@ -0,0 +1,95 @@
...
+        branches.add(caseTree.body());
+        if (caseTree.expression() == null) {
+          allBranchesArePresent = true;
+        }
+      }
+    }
+
+    private boolean allBranchesAreIdentical() {
+      Tree first = branches.get(0);
+      return branches.stream()
```
christophe-zurn-sonarsource: I think we should check that the list is not empty before accessing the element at index 0.
` when(x) {}` is valid kotlin code (not useful, but could still happen)

slang-checks/src/main/java/com/sonarsource/slang/checks/AllBranchesIdenticalCheck.java
```diff
@@ -0,0 +1,95 @@
...
+
+import static com.sonarsource.slang.checks.utils.SyntacticEquivalence.areEquivalent;
+
+@Rule(key = "S3923")
+public class AllBranchesIdenticalCheck implements SlangCheck {
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(IfTree.class, (ctx, tree) -> {
+      if (!(ctx.parent() instanceof IfTree)) {
```
christophe-zurn-sonarsource: Not sure, but what happens for nested ifs ?
`if(x)
  if(y)
    return true;
  else
    return true;
 else
  return false;
`
I think the nested one will not be considered since its parent is a `if` statement too

slang-checks/src/main/java/com/sonarsource/slang/checks/AllBranchesIdenticalCheck.java
pynicolas: Changed in efbbd33

slang-checks/src/main/java/com/sonarsource/slang/checks/AllBranchesIdenticalCheck.java
pynicolas: Fixed in efbbd33

PR #43, opened 2018-06-07
Fix syntactic equivalence for Kotlin operation expressions

pynicolas:


slang-kotlin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -207,13 +210,18 @@ private Tree createBinaryExpression(@NotNull KtBinaryExpression element, TreeMet
...
-      return new NativeTreeImpl(
-        metaData,
-        new KotlinNativeKind(element, element.getOperationReference().getReferencedNameElement().getText()),
-        Arrays.asList(leftOperand, rightOperand));
+      return createOperationExpression(element, metaData);
     }
   }
 
+  @NotNull
+  private Tree createOperationExpression(@NotNull PsiElement element, TreeMetaData metaData) {
```
christophe-zurn-sonarsource: I would prefer to have this method taking a `KtOperationExpression` element directly, that way we have the cast in the same place we test the type with `instanceof`, ensuring we do not misuse this method

pynicolas: Of course: 0e6e44a

PR #54, opened 2018-06-08
Fix FunctionDeclarationTree creation [SLANG] #44

christophe-zurn-sonarsource:


slang-antlr/src/test/java/com/sonarsource/slang/antlr/SLangConverterTest.java
```diff
@@ -67,7 +70,7 @@ public void testConverter() throws IOException {
 
     assertTree(tree).isInstanceOf(TopLevelTree.class).hasChildren(9);
     assertThat(numBinNodes.get()).isEqualTo(6);
-    assertThat(numIdentifierNode.get()).isEqualTo(18);
+    assertThat(numIdentifierNode.get()).isEqualTo(19);
```
pynicolas: We should do something about this test case, maybe just drop it.

slang-antlr/src/test/java/com/sonarsource/slang/antlr/SLangConverterTest.java
```diff
@@ -99,7 +102,12 @@ public void additive_expression_with_multiple_operands() {
   public void function() {
     FunctionDeclarationTree function = parseFunction("private int foo(x1, x2) { x1 + x2 }");
     assertThat(function.name().name()).isEqualTo("foo");
-    // assertTree(function.returnType()).isIdentifier("boolean");
+    NativeKind expectedModifierKind = new SNativeKind(SLangParser.MethodModifierContext.class, "private");
+    assertThat(function.modifiers()).hasSize(1);
+    Tree modifierTree = function.modifiers().get(0);
+    assertTree(modifierTree).isInstanceOf(NativeTree.class);
+    assertThat(((NativeTree) modifierTree).nativeKind()).isEqualTo(expectedModifierKind);
```
pynicolas: Why do we need that? I would prefer unit tests to show me why need different native kinds in some cases: in other words, its impact on syntactic equivalence. So far, I don't get why we need differentiatorObjs for slang.

christophe-zurn-sonarsource: @pynicolas I agree. But `SyntacticEquivalence` is in `slang-checks` module. So if it's fine with you, I'll move it into api module to do this?

slang-antlr/src/test/java/com/sonarsource/slang/antlr/SLangConverterTest.java
christophe-zurn-sonarsource: Ok, will drop it. It does not impact coverage on classes in `slang-antlr` module

slang-antlr/src/test/java/com/sonarsource/slang/antlr/SLangConverterTest.java
pynicolas: Fine for me. Although I'm not sure it's easy...

PR #59, opened 2018-06-11
SONARSLANG-1 Rule: Bad function name

andrea-guarino-sonarsource:


slang-checks/src/main/java/com/sonarsource/slang/checks/BadFunctionNameCheck.java
```diff
@@ -0,0 +1,55 @@
...
+import com.sonarsource.slang.checks.api.InitContext;
+import com.sonarsource.slang.checks.api.SlangCheck;
+import java.util.regex.Pattern;
+import org.sonar.check.Rule;
+import org.sonar.check.RuleProperty;
+
+@Rule(key = "S100")
+public class BadFunctionNameCheck implements SlangCheck {
+
+  private static final String DEFAULT_FORMAT = "^[a-z][a-zA-Z0-9]*$";
```
andrea-guarino-sonarsource: To be discussed, since default format depends on the analysed language.

sonar-slang-plugin/src/test/java/com/sonarsource/slang/kotlin/KotlinConverterTest.java
```diff
@@ -123,6 +123,14 @@ public void testFunctionDeclaration() {
     assertTree(emptyLambdaFunction.returnType()).isNull();
     assertThat(emptyLambdaFunction.formalParameters()).isEmpty();
     assertTree(emptyLambdaFunction.body()).isNull();
+
+    FunctionDeclarationTree extensionFunction = (FunctionDeclarationTree) kotlin("fun MutableList.swap(index1: Int, index2: Int) {}");
+    assertTree(extensionFunction.name()).isIdentifier("swap");
```
pynicolas: I don't think that we want that. It seems that the `MutableList` part is lost in the AST, which can be a big problem, at least for syntactic equivalence.

slang-checks/src/main/java/com/sonarsource/slang/checks/BadFunctionNameCheck.java
```diff
@@ -0,0 +1,57 @@
...
+  private String message(String name) {
+    return "Rename function " + name + " to match the regular expression " + format;
+  }
+
+  @Override
+  public void initialize(InitContext init) {
+    pattern = Pattern.compile(format);
+    init.register(FunctionDeclarationTree.class, (ctx, fnDeclarationTree) -> {
+      String name = fnDeclarationTree.name().name();
+      if (!pattern.matcher(name).matches()) {
```
pynicolas: The function name may be null, at least in Kotlin... Unfortunately, it seems we're not able to test that in slang textual representation...

slang-checks/src/main/java/com/sonarsource/slang/checks/CommonCheckList.java
```diff
@@ -35,7 +35,8 @@ private CommonCheckList() {
       IdenticalConditionsCheck.class,
       SelfAssignmentCheck.class,
       TodoCommentCheck.class,
-      TooManyParametersCheck.class);
+      TooManyParametersCheck.class,
+      BadFunctionNameCheck.class);
```
pynicolas: Let's keep class names ordered alphabetically.

Issue #47, opened 2018-06-08
Improve Kotlin parser readability


christophe-zurn-sonarsource:
`createElement` method can be improved to be more readable and maintanable

PR #63, opened 2018-06-11
Improve Kotlin converter #47

christophe-zurn-sonarsource:


sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -228,11 +204,71 @@ private Tree createFunctionDeclarationTree(@NotNull KtFunction functionElement,
...
+
+  private static Tree createIdentifierTree(TreeMetaData metaData, @Nullable String name) {
+    if (name == null) {
+      throw new ParseException("Identifier name cannot be null", metaData.textRange().start());
+    }
+    return new IdentifierTreeImpl(metaData, name);
+  }
+
+  private static Tree nullNative(TreeMetaData metaData) {
+    // We use the metadata of the parent here, as a NULL kind has no range itself
```
pynicolas: What's a "null native"? If it's a tree which does not exist, then I don't think we should represent it as an instance of Tree. I would prefer to represent the parent of such a tree as a native. For example, if Kotlin accepts an if statement with no "then" branch, then I think we should represent the whole statement with a NativeTree.

sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -114,88 +114,66 @@ public KotlinTreeVisitor(PsiFile psiFile, TreeMetaDataProvider metaDataProvider)
...
+      return createMatchCase(metaData, (KtWhenEntry) element);
+    } else if (isLiteral(element)) {
       return new LiteralTreeImpl(metaData, element.getText());
     } else if (element instanceof KtOperationExpression) {
-      return createOperationExpression((KtOperationExpression) element, metaData);
-    } else if (element instanceof KtParameter) {
-      return new IdentifierTreeImpl(metaData, ((KtParameter) element).getName());
+      return createOperationExpression(metaData, (KtOperationExpression) element);
+    } else if (element instanceof KtParameter && ((KtParameter) element).getName() != null) {
+      return createIdentifierTree(metaData, ((KtParameter) element).getName());
```
pynicolas: That seems wrong to me. A parameter is often more than an identifier, it can also have a type. Marcelo wants to add a `ParameterTree` to slang API for the purpose of his rule.

sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
christophe-zurn-sonarsource: Yes, I agree, this is a much better solution. Updated the PR accordingly

sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
christophe-zurn-sonarsource: Ok. I leave it as a native tree for now.

PR #64, opened 2018-06-12
Formal parameters

marcelo-sousa-sonarsource:


sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -171,7 +172,7 @@ private Tree createFunctionDeclarationTree(TreeMetaData metaData, KtFunction fun
     PsiElement nameIdentifier = functionElement.getNameIdentifier();
     Tree returnType = null;
     IdentifierTree identifierTree = null;
-    List<Tree> parametersList = list(functionElement.getValueParameters().stream());
+    List<ParameterTree> parametersList = (List<ParameterTree>)(List<?>)list(functionElement.getValueParameters().stream());
```
christophe-zurn-sonarsource: This should not work as a `ParameterTreeImpl` is never creted.
You can revert this change for now, and do another PR just related to kotlin AST convertion

slang-antlr/src/main/java/com/sonarsource/slang/parser/SLangConverter.java
```diff
@@ -305,7 +308,7 @@ private NativeTree simpleNativeTree(ParserRuleContext ctx) {
       return new NativeTreeImpl(meta(ctx), new SNativeKind(ctx, ctx.getText()), Collections.emptyList());
     }
 
-    private List<Tree> list(List<? extends ParseTree> rawChildren) {
+    private List list(List<? extends ParseTree> rawChildren) {
```
christophe-zurn-sonarsource: I don't like to have such a generic list as return type...
You could cast the returned list instead in the part that you need.
`resultList.stream().map(ParameterTree.class::cast).collect(toList());`

slang-api/src/test/java/com/sonarsource/slang/impl/ParameterTreeImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+    IdentifierTree identifierTreeX = new IdentifierTreeImpl(meta, "x");
+    IdentifierTree identifierTreeY = new IdentifierTreeImpl(meta, "y");
+    ParameterTreeImpl parameterTreeX = new ParameterTreeImpl(meta, identifierTreeX, null);
+    ParameterTreeImpl parameterTreeXTyped = new ParameterTreeImpl(meta, identifierTreeX, parameterType);
+    ParameterTreeImpl parameterTreeY = new ParameterTreeImpl(meta, identifierTreeY, parameterType);
+
+    assertThat(parameterTreeX.children()).hasSize(2);
+    assertThat(parameterTreeX.type()).isNull();
+    assertThat(parameterTreeX.identifier()).isEqualTo(identifierTreeX);
+    assertThat(areEquivalent(parameterTreeX, parameterTreeX)).isTrue();
```
christophe-zurn-sonarsource: I'd prefer to have another instance here `new IdentifierTreeImpl(meta, "x")` for the second tree, to ensure equivalence works on values of this new tree type and not only on references.

sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -160,8 +163,6 @@ private Tree convertElementToSlangAST(PsiElement element, TreeMetaData metaData)
     } else if (element instanceof KtDestructuringDeclarationEntry || isSimpleStringLiteralEntry(element)) {
       // To differentiate between the native trees of complex string template entries, we add the string value to the native kind
       return createNativeTree(metaData, new KotlinNativeKind(element, element.getText()), element);
-    } else if (element instanceof KtParameter) {
```
christophe-zurn-sonarsource: We can keep that part I believe. That way for normal loops (Ex: `for (a in container) {a}`) we keep information about `ParameterTree` for the variable. And we can add a test for this case

sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -229,16 +234,24 @@ private Tree createIfTree(TreeMetaData metaData, KtIfExpression element) {
     return new IfTreeImpl(metaData, condition, thenBranch, elseBranch);
   }
 
-  private Tree createParameter(TreeMetaData metaData, PsiElement element) {
-    String name = ((KtParameter) element).getName();
-    NativeKind kind;
+  private List<Tree> createParameterList(Stream<? extends KtParameter> stream) {
```
christophe-zurn-sonarsource: Actually, this method is not necessary since you check the list content in the parent caller method.
So keeping the `if (element instanceof KtParameter)` in the `convertElementToSlangAST` method and using `list(...)` would yield the same result

sonar-slang-plugin/src/test/java/com/sonarsource/slang/kotlin/KotlinConverterTest.java
```diff
@@ -30,16 +30,36 @@
...
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
 import static com.sonarsource.slang.testing.RangeAssert.assertRange;
 import static com.sonarsource.slang.testing.TreeAssert.assertTree;
 import static com.sonarsource.slang.testing.TreesAssert.assertTrees;
+import static com.sonarsource.slang.visitors.TreePrinter.tree2string;
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class KotlinConverterTest {
```
christophe-zurn-sonarsource: Tests related to functions should be updated with the parameters check.
You could add a `hasParameterNames(String... names)` assertion in testing module to simplify your life :)

slang-testing/src/main/java/com/sonarsource/slang/testing/TreeAssert.java
```diff
@@ -49,6 +50,14 @@ public TreeAssert isIdentifier(String expectedName) {
     return this;
   }
 
+  public TreeAssert hasParameterName(String expectedIdentifierName) {
```
christophe-zurn-sonarsource: No unit tests were added for this assertion

sonar-slang-plugin/src/test/java/com/sonarsource/slang/kotlin/KotlinConverterTest.java
```diff
@@ -140,13 +125,15 @@ public void testFunctionDeclaration() {
...
     assertThat(constructorFunction.formalParameters()).hasSize(2);
+    assertTree(constructorFunction).hasParameterNames("a", "b");
     assertTree(constructorFunction.body()).isNull();
 
     FunctionDeclarationTree emptyLambdaFunction = (FunctionDeclarationTree) kotlin("{ }");
     assertTree(emptyLambdaFunction.name()).isNull();
     assertThat(emptyLambdaFunction.modifiers()).isEmpty();
     assertTree(emptyLambdaFunction.returnType()).isNull();
     assertThat(emptyLambdaFunction.formalParameters()).isEmpty();
+    assertTree(functionWithPrivate).hasParameterNames();
```
christophe-zurn-sonarsource: `functionWithPrivate` is not the correct variable here. Also, this assert will not do anything special

sonar-slang-plugin/src/test/java/com/sonarsource/slang/kotlin/KotlinConverterTest.java
```diff
@@ -124,13 +105,17 @@ public void testFunctionDeclaration() {
...
 
-    FunctionDeclarationTree functionWithInternalModifier = (FunctionDeclarationTree) kotlin("internal fun function1(a: Int, b: String): Boolean = true");
+    FunctionDeclarationTree functionWithInternalModifier = (FunctionDeclarationTree) kotlin("internal fun function1(a: Int, c: String): Boolean = true");
     assertTree(functionWithInternalModifier.body()).isNotNull();
     assertThat(functionWithInternalModifier.modifiers()).hasSize(1);
+    assertTree(functionWithInternalModifier).hasParameterNames("a", "c");
 
     FunctionDeclarationTree functionWithPrivate = (FunctionDeclarationTree) kotlin("private fun function2() {}");
+    assertThat(functionWithPrivate.formalParameters()).isEmpty();
+    assertTree(functionWithPrivate).hasParameterNames();
```
christophe-zurn-sonarsource: This assert `.hasParameterNames();` does not check anything special here. Since we already know `functionWithPrivate` is of type `FunctionDeclarationTree`

slang-testing/src/test/java/com/sonarsource/slang/testing/TreeAssertTest.java
```diff
@@ -65,6 +67,11 @@ public void not_an_identifier() {
     assertTree(LITERAL_42).isIdentifier("abc");
   }
 
+  @Test
+  public void parameter_has_identifier() {
```
christophe-zurn-sonarsource: For quality gate, we should also cover the case that the assert fails

slang-testing/src/main/java/com/sonarsource/slang/testing/TreeAssert.java
```diff
@@ -58,6 +61,21 @@ public TreeAssert hasParameterName(String expectedIdentifierName) {
     return this;
   }
 
+  public TreeAssert hasParameterNames(String... names) {
```
christophe-zurn-sonarsource: Unit tests are missing for this new assertion

PR #66, opened 2018-06-12
Conversion of unary expressions to SLang AST

marcelo-sousa-sonarsource:


slang-api/src/main/java/com/sonarsource/slang/impl/UnaryExpressionTreeImpl.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.sonarsource.slang.api.Tree;
+import com.sonarsource.slang.api.TreeMetaData;
+import com.sonarsource.slang.api.UnaryExpressionTree;
+import java.util.Collections;
+import java.util.List;
+
+public class UnaryExpressionTreeImpl extends BaseTreeImpl implements UnaryExpressionTree {
+
+  private final UnaryExpressionTree.Operator operator;
```
christophe-zurn-sonarsource: `UnaryExpressionTree.` part is not necessary here

slang-api/src/main/java/com/sonarsource/slang/utils/SyntacticEquivalence.java
```diff
@@ -67,16 +68,13 @@ public static boolean areEquivalent(@Nullable Tree first, @Nullable Tree second)
...
-      }
-    } else if (first instanceof AssignmentExpressionTree
-      && (((AssignmentExpressionTree) first).operator() != ((AssignmentExpressionTree) second).operator())) {
+    } else if ((first instanceof NativeTree) && (!((NativeTree) first).nativeKind().equals(((NativeTree) second).nativeKind()))) {
+      return false;
+    } else if ((first instanceof UnaryExpressionTree) && ((UnaryExpressionTree) first).operator() != ((UnaryExpressionTree) second).operator()) {
+      return false;
+    } else if ((first instanceof BinaryExpressionTree) && (((BinaryExpressionTree) first).operator() != ((BinaryExpressionTree) second).operator())) {
+      return false;
+    } else if ((first instanceof AssignmentExpressionTree) && (((AssignmentExpressionTree) first).operator() != ((AssignmentExpressionTree) second).operator())) {
```
christophe-zurn-sonarsource: I think this could be refactored a bit. You could put all this conditions in a function (for instance `hasDifferentFields(first, second)` ). You would then only have 1  branch for all these `return false` cases

slang-api/src/test/java/com/sonarsource/slang/impl/UnaryExpressionTreeImplTest.java
```diff
@@ -0,0 +1,50 @@
...
+  @Test
+  public void test() {
+    TreeMetaData meta = null;
+    Tree condition = new IdentifierTreeImpl(meta, "x");
+    Tree negCondition = new UnaryExpressionTreeImpl(meta, UnaryExpressionTree.Operator.NEGATE, condition);
+    Tree nativeTree = new NativeTreeImpl(meta, new TypeNativeKind(), Arrays.asList(condition));
+    Tree negNative = new UnaryExpressionTreeImpl(meta, UnaryExpressionTree.Operator.NEGATE, nativeTree);
+
+    assertThat(negCondition.children()).containsExactly(condition);
+    assertThat(areEquivalent(negCondition, negCondition)).isTrue();
```
christophe-zurn-sonarsource: You can use `assertTree(x).isEquivalentTo(y)` instead of testing the resulting boolean.
Also, I would prefer to have another instance of `negCondition` for testing equivalence, as this case will only test the `first == second` in the `areEquivalent` method

slang-api/src/main/java/com/sonarsource/slang/api/UnaryExpressionTree.java
```diff
@@ -0,0 +1,32 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.sonarsource.slang.api;
+
+public interface UnaryExpressionTree extends Tree {
+
+  public enum Operator {
+    NEGATE
```
christophe-zurn-sonarsource: I don't know if there's a naming convention among analyzers, but I don't believe I've seen this name for this operator before.

slang-antlr/src/main/antlr4/com/sonarsource/slang/parser/SLang.g4
```diff
@@ -185,6 +190,10 @@ assignmentOperator
   :  '=' | '+=' | '-=' | '*=' | '%='
   ;
 
+unaryOperator
+  :  '!'
```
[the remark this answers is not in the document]
marcelo-sousa-sonarsource: Lets not add more operators in this PR.

slang-antlr/src/main/java/com/sonarsource/slang/parser/SLangConverter.java
```diff
@@ -140,6 +143,11 @@ public Tree visitNativeExpression(SLangParser.NativeExpressionContext ctx) {
       return nativeTree(ctx, ctx.nativeBlock());
     }
 
+    @Override
+    public Tree visitParenthesizedExpression(SLangParser.ParenthesizedExpressionContext ctx) {
+      return visit(ctx.statementOrExpression());
```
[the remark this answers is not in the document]
marcelo-sousa-sonarsource: Yes, we need this visitor; otherwise the default visitor will iterate over all children and return null. As a consequence, the unit test will fail.

slang-api/src/main/java/com/sonarsource/slang/api/UnaryExpressionTree.java
marcelo-sousa-sonarsource: Lets leave it for now.

slang-api/src/test/java/com/sonarsource/slang/impl/UnaryExpressionTreeImplTest.java
marcelo-sousa-sonarsource: We cannot use assertTree here because it would create a cycle in the dependencies.

slang-antlr/src/main/java/com/sonarsource/slang/parser/SLangConverter.java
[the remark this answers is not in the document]
christophe-zurn-sonarsource: Mmm ok. But I think the discussion about whether to have `ParenthesesExpression` was still a open. But it's true that it does not concern this PR. But it need to be kept in mind

PR #67, opened 2018-06-12
Templated string literals with identifiers or blocks are now considered native elements

christophe-zurn-sonarsource:


sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -330,8 +331,18 @@ private static boolean isError(PsiElement element) {
...
+
+  private static boolean isSimpleStringLiteral(PsiElement element) {
+    return element instanceof KtStringTemplateExpression
+      && Stream.of(((KtStringTemplateExpression) element).getEntries()).allMatch(KotlinTreeVisitor::isSimpleStringLiteralEntry);
+  }
+
+  private static boolean isSimpleStringLiteralEntry(PsiElement element) {
+    return element instanceof KtLiteralStringTemplateEntry
       || element instanceof KtEscapeStringTemplateEntry
       || (element instanceof KtStringTemplateExpression && !((KtStringTemplateExpression) element).hasInterpolation());
```
pynicolas: I must say that I don't understand everything, but removing this line does not impact unit tests...

sonar-slang-plugin/src/test/java/com/sonarsource/slang/kotlin/KotlinConverterTest.java
```diff
@@ -163,24 +163,28 @@ public void testSimpleStringLiterals() {
...
   @Test
   public void testStringWithBlock() {
     Tree stringWithBlock = kotlinStatement("\"block ${1 == 1}\"");
-    assertTree(stringWithBlock).isInstanceOf(NativeTree.class);
-    assertTree(stringWithBlock.children().get(0)).isLiteral("block ");
+    assertTree(stringWithBlock).isInstanceOf(NativeTree.class).hasChildren(NativeTree.class, NativeTree.class);
     Tree blockExpressionContainer = stringWithBlock.children().get(1);
     assertTree(blockExpressionContainer).isInstanceOf(NativeTree.class);
     assertThat(blockExpressionContainer.children()).hasSize(1);
     assertTree(blockExpressionContainer.children().get(0)).isBinaryExpression(BinaryExpressionTree.Operator.EQUAL_TO);
```
pynicolas: Being able to write assertions on descendants easily would help in such cases.

christophe-zurn-sonarsource: I agree. Created a related Jira task for it: https://jira.sonarsource.com/browse/SONARSLANG-25

sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
christophe-zurn-sonarsource: Right. This was a mistake on my part during refactoring.
Small explanation: the part `(element instanceof KtStringTemplateExpression && !((KtStringTemplateExpression) element).hasInterpolation());` does the same thing as `Stream.of(((KtStringTemplateExpression) element).getEntries()).allMatch(KotlinTreeVisitor::isSimpleStringLiteralEntry);` and it is not needed in that method. 
As here we only check the entries of the String template that are string literals (i.e. we still want to differentiate in the native `StringTemplate` element its children).
Updated it before merge

PR #69, opened 2018-06-12
Automated integration tests

pynicolas:


pynicolas:
@christophe-zurn-sonarsource It can happen that 'sonar-slang-plugin.jar is not found: in that case, orchestrator fails with a clear message (if I remember correctly).

PR #74, opened 2018-06-13
Add descendants to API

christophe-zurn-sonarsource:


slang-api/src/test/java/com/sonarsource/slang/api/TreeTest.java
```diff
@@ -0,0 +1,58 @@
...
+import static com.sonarsource.slang.TreeCreationUtils.literal;
+import static com.sonarsource.slang.TreeCreationUtils.simpleFunction;
+import static com.sonarsource.slang.TreeCreationUtils.topLevel;
+import static com.sonarsource.slang.api.BinaryExpressionTree.Operator.EQUAL_TO;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TreeTest {
+
+  @Test
+  public void test() {
```
marcelo-sousa-sonarsource: Add a native child in the test to get a bit more coverage.

slang-testing/src/main/java/com/sonarsource/slang/testing/TreeAssert.java
```diff
@@ -97,6 +97,24 @@ public TreeAssert hasChildren(Class... classes) {
     return this;
   }
 
+  public TreeAssert hasDescendant(Tree expected) {
```
marcelo-sousa-sonarsource: Refactor this method and hasNotDescendant using an auxiliary method that returns whether a tree is a descendant. That logic is replicated in both methods and it could be good to isolate in case it changes in the future.

slang-testing/src/test/java/com/sonarsource/slang/testing/TreeAssertTest.java
```diff
@@ -163,6 +165,26 @@ public void notequivalent_failure() {
...
+  public void hasdescendant_ok() {
+    assertTree(ABC_PLUS_ABC_PLUS_42).hasDescendant(new LiteralTreeImpl(null, "42"));
+  }
+
+  @Test(expected = AssertionError.class)
+  public void hasdescendant_failure() {
+    assertTree(ABC_PLUS_ABC_PLUS_42).hasDescendant(new LiteralTreeImpl(null, "43"));
+  }
+
+  @Test
```
marcelo-sousa-sonarsource: I fail to see the value of these tests. It is clear from the implementation that a tree cannot be both a descendant and not a descendant. If you really want to keep these tests, why not just assert !hasDescendant == hasNotDescendant?

christophe-zurn-sonarsource: I agree, it's clear what they do from the implementation, but they still need to be unit tested :)
I don't think testing `!hasDescendant == hasNotDescendant` would be useful, even less now that they share common method implementation. 
Here, we at least ensure that the assert works as we expect

slang-api/src/main/java/com/sonarsource/slang/api/Tree.java
```diff
@@ -20,11 +20,19 @@
...
 public interface Tree {
 
   List<Tree> children();
 
   TreeMetaData metaData();
 
+  default Stream<Tree> descendants() {
+    return children().stream()
+      .filter(Objects::nonNull)
+      .flatMap(tree -> children().isEmpty() ? Stream.of(tree) : Stream.concat(Stream.of(tree), tree.descendants()));
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: Oh, actually the children check was unecessary, nice catch!
Personally I'd rather keep the concat stream version:
`.flatMap(tree -> Stream.concat(Stream.of(tree), tree.descendants()))`
`.flatMap(tree -> Stream.of(Stream.of(tree), tree.descendants()).flatMap(x -> x))`
I think it is a bit more readable than a nested flatMap

slang-api/src/test/java/com/sonarsource/slang/api/TreeTest.java
```diff
@@ -23,32 +23,38 @@
...
+import static com.sonarsource.slang.utils.TreeCreationUtils.literal;
+import static com.sonarsource.slang.utils.TreeCreationUtils.simpleFunction;
+import static com.sonarsource.slang.utils.TreeCreationUtils.simpleNative;
+import static com.sonarsource.slang.utils.TreeCreationUtils.topLevel;
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class TreeTest {
 
+  private static final NativeKind SIMPLE_KIND = new NativeKind() {
+  };
```
marcelo-sousa-sonarsource: Refactor to have NativeKind () { };

PR #75, opened 2018-06-13
Add Rule: Collapsible If Statements

andrea-guarino-sonarsource:


slang-checks/src/main/java/com/sonarsource/slang/checks/CollapsibleIfStatementsCheck.java
```diff
@@ -0,0 +1,63 @@
...
+
+@Rule(key = "S1066")
+public class CollapsibleIfStatementsCheck implements SlangCheck {
+  private static final String MESSAGE = "Merge this if statement with the nested one.";
+  private static final String SECONDARY_MESSAGE = "Nested \"if\" statement";
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(IfTree.class, (ctx, ifTreeStatement) -> Optional.of(ifTreeStatement)
+      .filter(ifTree -> ifTree.elseBranch() == null)
```
christophe-zurn-sonarsource: I don't think this filter operation is necessary. As the next method will return an empty optional if that's the case

sonar-slang-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S1066.html
```diff
@@ -0,0 +1 @@
+<p>Merging collapsible <code>if</code> statements increases the code's readability.</p>
```
christophe-zurn-sonarsource: The Compliant / Noncompliant sections are missing. You can create a `Kotlin` subtask in the RSPEC to override these parts

slang-checks/src/test/resources/com/sonarsource/slang/checks/CollapsibleIfStatements.slang
```diff
@@ -0,0 +1,57 @@
...
+  }
+} else {
+}
+
+if (a)
+  if (b) {
+  } else {
+  }
+
+if (a) { // Noncompliant {{Merge this if statement with the nested one.}}
```
christophe-zurn-sonarsource: You could add a location check on 1 of these noncompliant example

slang-checks/src/main/java/com/sonarsource/slang/checks/CollapsibleIfStatementsCheck.java
```diff
@@ -0,0 +1,63 @@
...
+import org.sonar.check.Rule;
+
+@Rule(key = "S1066")
+public class CollapsibleIfStatementsCheck implements SlangCheck {
+  private static final String MESSAGE = "Merge this if statement with the nested one.";
+  private static final String SECONDARY_MESSAGE = "Nested \"if\" statement";
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(IfTree.class, (ctx, ifTreeStatement) -> Optional.of(ifTreeStatement)
```
christophe-zurn-sonarsource: Is there a special reason you used an `Optional` here?
It's not wrong, but it seems strange to me to use it like this on an object we know exists...

slang-checks/src/main/java/com/sonarsource/slang/checks/CollapsibleIfStatementsCheck.java
andrea-guarino-sonarsource: I think this check on "elseBranch" is still necessary because the next method will receive only the "thenBranch".

slang-checks/src/main/java/com/sonarsource/slang/checks/CollapsibleIfStatementsCheck.java
andrea-guarino-sonarsource: I like to have this fluent api notation, but I agree it's a bit weird in this case. :)
I will remove the `Optional.of` wrapper.

slang-checks/src/main/java/com/sonarsource/slang/checks/CollapsibleIfStatementsCheck.java
christophe-zurn-sonarsource: You're right. I misread that part

PR #76, opened 2018-06-13
SONARSLANG-11  Rule: String literals should not be duplicated

christophe-zurn-sonarsource:


slang-checks/src/test/resources/com/sonarsource/slang/checks/StringLiteralDuplicated.threshold_4.slang
```diff
@@ -0,0 +1,31 @@
+x = "string literal1"                           // Compliant - only appears 3 times which is less than the custom threshold of 4
```
pynicolas: I would prefer that this file does not mostly duplicate StringLiteralDuplicated.slang.
I think it should only be used to test the threshold.

sonar-slang-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S1192.json
```diff
@@ -0,0 +1,18 @@
+{
+  "title": "String literals should not be duplicated",
+  "type": "CODE_SMELL",
+  "status": "ready",
+  "remediation": {
+    "func": "Linear with offset",
+    "linearDesc": "per duplicate instance",
+    "linearOffset": "2min",
+    "linearFactor": "2min"
```
pynicolas: That's not consistent with the issues we create: we should report a "gap" which will be used to compute the issue debt.

PR #77, opened 2018-06-14
SONARSLANG-7 Rule: Two branches in a conditional structure should not…

pynicolas:
… have exactly the same implementation


slang-checks/src/test/resources/com/sonarsource/slang/checks/DuplicateBranch.slang
```diff
@@ -0,0 +1,75 @@
...
+    foo;
+  } else {
+    foo;
+    foo;
+  }
+
+  if (x) {
+    foo;
+    foo;
+  } else if (y) { // Noncompliant {{This branch's code block is the same as the block for the branch on line 22.}}
```
andrea-guarino-sonarsource: Could you please add a location check on one of those noncompliant example?

pynicolas: Done in 2e5ddad.

PR #78, opened 2018-06-14
SONARSLANG-8 Rule: no nested match

andrea-guarino-sonarsource:


slang-api/src/test/java/com/sonarsource/slang/impl/MatchTreeImplTest.java
```diff
@@ -35,7 +36,8 @@ public void test() {
     TreeMetaData meta = null;
     Tree expression = new IdentifierTreeImpl(null, "x");
     MatchCaseTree case1 = new MatchCaseTreeImpl(null, null, new LiteralTreeImpl(meta, "42"));
-    MatchTree tree = new MatchTreeImpl(meta, expression, Collections.singletonList(case1));
+    Token keywordToken = new TokenImpl(new TextRangeImpl(1,0,1,20), "match", true);
+    MatchTree tree = new MatchTreeImpl(meta, expression, Collections.singletonList(case1), keywordToken);
     assertThat(tree.children()).containsExactly(expression, case1);
     assertThat(tree.expression()).isEqualTo(expression);
     assertThat(tree.cases()).containsExactly(case1);
```
christophe-zurn-sonarsource: just for completeness sake, we could add an assert on keyword in this unit test too

slang-checks/src/main/java/com/sonarsource/slang/checks/NestedMatchCheck.java
```diff
@@ -0,0 +1,54 @@
...
+          ctx.reportIssue(match.keyword().textRange(), MessageFormat.format(MESSAGE, match.keyword().text()));
+        }
+      ));
+  }
+
+  private static Optional<Tree> getNestedMatchTrees(MatchCaseTree matchCaseTree) {
+    return matchCaseTree
+      .descendants()
+      .filter(node -> node instanceof MatchTree)
+      .findFirst();
```
christophe-zurn-sonarsource: I would return all nested `MatchTree` descendants instead of only the first one, then you could use `flatMap` and `distinct` on the returned stream to get single `MatchTree` for each violation.
Otherwise we might miss issue in the following case
```
when(x) {
  1 -> {
    when(y) { // OK -> issue raised correctly
      ...
    }
    when(z) { // KO -> no issue raised here
      ...
    }
  }
}
```

sonar-slang-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S1821.html
```diff
@@ -0,0 +1,34 @@
+<p>Nested <code>switch</code> structures are difficult to understand because you can easily confuse the cases of an inner <code>switch</code> as
```
christophe-zurn-sonarsource: You should also update description in Kotlin RSPEC subtask (`switch` -> `when`)

sonar-slang-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S1821.json
```diff
@@ -0,0 +1,16 @@
+{
+  "title": "\"switch\" statements should not be nested",
```
christophe-zurn-sonarsource: You should also update rule title in  Kotlin RSPEC subtask (`switch` -> `when`)

slang-checks/src/main/java/com/sonarsource/slang/checks/NestedMatchCheck.java
andrea-guarino-sonarsource: Well spotted! I assumed each case being an expression, but that's not the case.

sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -273,7 +276,17 @@ private Tree createMatchTree(TreeMetaData metaData, KtWhenExpression element) {
       subjectExpression,
       whenExpressions.stream()
         .map(MatchCaseTree.class::cast)
-        .collect(Collectors.toList()));
+        .collect(Collectors.toList()),
+      getWhenKeyword(element));
+  }
+
+  @NotNull
```
christophe-zurn-sonarsource: This annotation is not necessary as by default we have `@javax.annotation.ParametersAreNonnullByDefault` on all packages. So a warning would be raised even without it

slang-antlr/src/main/java/com/sonarsource/slang/parser/SLangConverter.java
```diff
@@ -252,7 +252,18 @@ public Tree visitMatchExpression(SLangParser.MatchExpressionContext ctx) {
...
+      return new MatchTreeImpl(
+        meta,
+        visit(ctx.statementOrExpression()),
+        cases,
+        getMatchKeyword(ctx.MATCH().getSymbol()));
+    }
+
+    private static com.sonarsource.slang.api.Token getMatchKeyword(Token matchToken) {
+      TokenLocation location = new TokenLocation(matchToken.getLine(), matchToken.getCharPositionInLine(), matchToken.getText());
+      TextRange textRange = new TextRangeImpl(location.startLine(), location.startLineOffset(), location.endLine(), location.endLineOffset());
```
christophe-zurn-sonarsource: These 2 lines could be put in a helper function since it is used the same way on line 98-99

andrea-guarino-sonarsource: Indeed, I created a method `toSlangToken` as well in another PR (for If-then-else keywords)
https://github.com/SonarSource/slang/pull/86/files#diff-44906d002be9e3d0bff18e98b3b954c3R270
I will copy it here as well

PR #79, opened 2018-06-14
SONARSLANG-2 Rule: Functions should not have identical implementations

christophe-zurn-sonarsource:


slang-checks/src/main/java/com/sonarsource/slang/checks/DuplicatedFunctionImplementationCheck.java
```diff
@@ -0,0 +1,108 @@
...
+    });
+
+  }
+
+  private static boolean hasMinimumLineNumber(FunctionDeclarationTree function) {
+    BlockTree functionBody = function.body();
+    if (functionBody == null) {
+      return false;
+    }
+    // TODO replace by > 2 LoC when available
```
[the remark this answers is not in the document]
pynicolas: The original RSPEC defines the exception on the number of statements, not on the number of lines.

slang-checks/src/main/java/com/sonarsource/slang/checks/DuplicatedFunctionImplementationCheck.java
```diff
@@ -0,0 +1,108 @@
...
+  private static void reportDuplicate(CheckContext ctx, FunctionDeclarationTree original, FunctionDeclarationTree duplicate) {
+    IdentifierTree identifier = original.name();
+    String message;
+    Tree secondaryTree;
+    if (identifier != null) {
+      secondaryTree = identifier;
+      message = String.format(MESSAGE, identifier.name(), original.metaData().textRange().start().line());
+    } else {
+      secondaryTree = original;
+      message = String.format(MESSAGE_NO_NAME, original.metaData().textRange().start().line());
```
pynicolas: Minor: you could factorize `original.metaData().textRange().start().line()` which is also used 3 lines above.

pynicolas:
I think that we currently raise too many issues: 3000 issues on the ruling is a lot for this rule. I believe that changing the threshold from number of lines to number of statements can decrease that number.

christophe-zurn-sonarsource:
@pynicolas Yes that was what I was thinking would happen when changing lines to LoC. Now I updated to implementation to raise issue when there is at least 2 statements in the body of the function.
It reduced the number of issue by around half, but ~1500 issues still seem like a lot to me.

pynicolas:
To me, the rule still seems noisy, even when excluding test files from the results of the ruling (`grep kotlin its/ruling/src/test/resources/expected/kotlin-S4144.json | grep -i -v test`).
Example: QueryCriteria.kt
Suggestion: what about searching duplicates among functions which have the same parent tree? Most often, that would be a class, but we don't really need to know whether it's a class.

christophe-zurn-sonarsource:
@pynicolas Ok. I think this is actually a good suggestion. That way it will still compare method of the same class, as well as global functions and lambda inside the same function/methods.

slang-checks/src/main/java/com/sonarsource/slang/checks/DuplicatedFunctionImplementationCheck.java
```diff
@@ -0,0 +1,116 @@
...
+        .filter(f -> areDuplicatedImplementation(original, f))
+        .forEach(duplicate -> {
+          reportDuplicate(ctx, original, duplicate);
+          reportedDuplicates.add(duplicate);
+        });
+    });
+
+  }
+
+  private static boolean hasMinimumLineNumber(FunctionDeclarationTree function) {
```
pynicolas: Could be renamed to something like "hasMinimumSize".

slang-checks/src/main/java/com/sonarsource/slang/checks/DuplicatedFunctionImplementationCheck.java
```diff
@@ -47,12 +51,17 @@
   @Override
   public void initialize(InitContext init) {
     init.register(TopLevelTree.class, (ctx, tree) -> {
-      List<FunctionDeclarationTree> functionDeclarations = tree.descendants()
-        .filter(FunctionDeclarationTree.class::isInstance)
-        .map(FunctionDeclarationTree.class::cast)
-        .collect(Collectors.toList());
+      Map<Tree, List<FunctionDeclarationTree>> functionsByParents = new HashMap<>();
```
pynicolas: A different implementation could be to make the check register on `Tree.class` and check its children. I'm not sure whether it would be better.

christophe-zurn-sonarsource: Yes, I thought about this as well. I wasn't sure which to go with. In the end I chose this one to avoid using `instanceof` in the rule.

slang-checks/src/main/java/com/sonarsource/slang/checks/DuplicatedFunctionImplementationCheck.java
christophe-zurn-sonarsource: Done in 2f63a47c9213ef9997ffb53c8cd9ebeec8b373da

PR #80, opened 2018-06-15
SONARSLANG-26 Rule: Track uses of "FIXME" tags

christophe-zurn-sonarsource:


slang-checks/src/main/java/com/sonarsource/slang/checks/FixMeCommentCheck.java
```diff
@@ -0,0 +1,54 @@
...
+
+  private final Pattern fixMePattern = Pattern.compile("(?i)(^|[^\\p{L}])(fixme)");
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(TopLevelTree.class, (ctx, tree) -> tree.allComments().forEach(comment -> {
+      Matcher matcher = fixMePattern.matcher(comment.textWithDelimiters());
+      if (matcher.find()) {
+        TextPointer start = comment.textRange().start();
+        TokenLocation location = new TokenLocation(start.line(), start.lineOffset(), comment.textWithDelimiters().substring(0, matcher.start(2)));
```
andrea-guarino-sonarsource: Can you please split this line?

slang-checks/src/main/java/com/sonarsource/slang/checks/FixMeCommentCheck.java
```diff
@@ -0,0 +1,54 @@
...
+        TextRange fixMeRange = new TextRangeImpl(
+          new TextPointerImpl(location.endLine(), location.endLineOffset()),
+          new TextPointerImpl(location.endLine(), location.endLineOffset() + 5));
+        ctx.reportIssue(fixMeRange, "Take the required action to fix the issue indicated by this \"FIXME\" comment.");
```
andrea-guarino-sonarsource: Please, move this message in a dedicated static property.

pynicolas: To me, it can stay as it is. I don't mind defining issue messages with constants, but I don't think we should mandate that.

PR #82, opened 2018-06-15
SONARSLANG-4 Rule: Unused Function Parameter

marcelo-sousa-sonarsource:


slang-checks/src/main/java/com/sonarsource/slang/checks/UnusedFunctionParameterCheck.java
```diff
@@ -0,0 +1,77 @@
...
+public class UnusedFunctionParameterCheck implements SlangCheck {
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(FunctionDeclarationTree.class, (ctx, functionDeclarationTree) -> {
+      if (!(ctx.parent() instanceof TopLevelTreeImpl)) {
+        return;
+      }
+
+      if (functionDeclarationTree.body() == null || functionDeclarationTree.body().descendants() == null) {
```
christophe-zurn-sonarsource: You could merge this if with the previous one for conciseness. Also `descendants()` cannot be null

slang-checks/src/main/java/com/sonarsource/slang/checks/UnusedFunctionParameterCheck.java
```diff
@@ -0,0 +1,77 @@
...
+      ParameterTree firstUnused = unusedParameters.get(0);
+      String msg;
+
+      if (unusedParameters.size() > 1) {
+        msg = "Remove these unused function parameters.";
+      } else {
+        msg = "Remove this unused function parameter \"" + firstUnused.identifier().name() + "\".";
+      }
+
+      ctx.reportIssue(firstUnused, msg, secondaryLocations);
```
christophe-zurn-sonarsource: The rule implementation in other analyzers highlight only the identifiers. I think we should do the same here

its/ruling/pom.xml
```diff
@@ -27,5 +27,10 @@
       <artifactId>sonar-analyzer-commons</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.sonarsource.sonar-lits-plugin</groupId>
+      <artifactId>sonar-lits-plugin</artifactId>
```
christophe-zurn-sonarsource: Why this new dependency?

slang-checks/src/main/java/com/sonarsource/slang/checks/UnusedFunctionParameterCheck.java
```diff
@@ -0,0 +1,77 @@
...
+
+import static com.sonarsource.slang.utils.SyntacticEquivalence.areEquivalent;
+
+@Rule(key = "S1172")
+public class UnusedFunctionParameterCheck implements SlangCheck {
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(FunctionDeclarationTree.class, (ctx, functionDeclarationTree) -> {
+      if (!(ctx.parent() instanceof TopLevelTreeImpl)) {
```
christophe-zurn-sonarsource: That's an example of False Negative we might want to track. Only global functions will raise an issue for now. Class methods and lambda will never raise issue for now. You could create a jira ticket related to this

slang-checks/src/test/resources/com/sonarsource/slang/checks/UnusedFunctionParameter.slang
```diff
@@ -0,0 +1,22 @@
...
+
+void fun fooBar(int a, int b) {}  // Noncompliant {{Remove these unused function parameters.}}
+//              ^^^^^  ^^^^^<
+
+void fun fooBar(int a, int b) {   // Noncompliant {{Remove this unused function parameter "b".}}
+//                     ^^^^^
+  x = a;
+}
+
+void fun fooBar(int a, int b) {   // Noncompliant {{Remove this unused function parameter "a".}}
```
christophe-zurn-sonarsource: We could have one of these test with a parameter without type to ensure it works in that case too

slang-checks/src/main/java/com/sonarsource/slang/checks/UnusedFunctionParameterCheck.java
```diff
@@ -0,0 +1,77 @@
...
+          .map(SecondaryLocation::new)
+          .collect(Collectors.toList());
+
+      ParameterTree firstUnused = unusedParameters.get(0);
+      String msg;
+
+      if (unusedParameters.size() > 1) {
+        msg = "Remove these unused function parameters.";
+      } else {
+        msg = "Remove this unused function parameter \"" + firstUnused.identifier().name() + "\".";
```
christophe-zurn-sonarsource: In java analyzer, they always have this message, and each secondary location also contains this message.
Cf: https://sonarcloud.io/project/issues?id=org.apache.tika%3Atika&issues=8246e5b6-acb7-490e-b1ad-7dd56827218c&open=8246e5b6-acb7-490e-b1ad-7dd56827218c

its/ruling/pom.xml
marcelo-sousa-sonarsource: @pynicolas suggested we added this dependency to debug ITs locally.

christophe-zurn-sonarsource: Ah ok :+1:

pynicolas: Orchestrator now looks into the local Maven repository to find the lits plugin. This new dependency forces Maven to download it before running the tests. Potential problem: it has to be the same version as used in the test.

PR #84, opened 2018-06-16
SONARSLANG-3 Syntax highlighting

pynicolas:


pynicolas:
Too complex. #85 is simpler.

PR #85, opened 2018-06-16
SONARSLANG-3 Syntax highlighting

pynicolas:


slang-api/src/main/java/com/sonarsource/slang/impl/TreeMetaDataProvider.java
```diff
@@ -61,5 +64,13 @@ public TextRange textRange() {
         .filter(comment -> comment.textRange().isInside(textRange))
         .collect(Collectors.toList());
     }
+
+    @Override
+    public List<Token> tokens() {
+      // TODO improve performance by storing an ordered list of tokens
```
christophe-zurn-sonarsource: Another possible improvement would be to have the tokens/comments computed only once and stored in the `TreeMetaDataImpl` so that it doesn't recompute it when called multiple times

sonar-slang-plugin/src/test/java/com/sonarsource/slang/plugin/SyntaxHighlighterTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+  @Rule
+  public TemporaryFolder tempFolder = new TemporaryFolder();
+
+  @Before
+  public void setUp() {
+    sensorContext = SensorContextTester.create(tempFolder.getRoot());
+  }
+
+  private void highlight(String sqlCode) throws IOException {
```
christophe-zurn-sonarsource: `sqlCode` ?
I have a good hint on which analyzer you inspired from for this part :)

PR #86, opened 2018-06-18
SONARSLANG-30 Rule: "if...else if" constructs should end with "else" clauses

andrea-guarino-sonarsource:


slang-api/src/main/java/com/sonarsource/slang/api/ConditionalKeyword.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.sonarsource.slang.api;
+
+import javax.annotation.CheckForNull;
+
+public interface ConditionalKeyword {
```
christophe-zurn-sonarsource: I don't really like this `ConditionalKeyword` wrapper. I don't really see the use and I find it strange to have `MatchTree` returning a `Token` when calling `keyword` method, and here the same method returns a completely different object.
I'd prefer to have the keywords directly in the `Tree` API to be consistent. And for now we don't need to add this `thenKeyword` since we have no use case so far.
An alternative could be to separate the else part into another Tree as they do in some of the other analyzers (`ElseClauseTree`) to have consistency between the `else` keyword and branch tree

andrea-guarino-sonarsource: I see your point about consistency and I agree. As discussed during the standup, I'm going to add an ElseClauseTree which implements a keyword method.

christophe-zurn-sonarsource:
LGTM

PR #88, opened 2018-06-18
SONARSLANG-35 Exception handling in slang

marcelo-sousa-sonarsource:


slang-antlr/src/main/java/com/sonarsource/slang/parser/SLangConverter.java
```diff
@@ -262,6 +265,24 @@ public Tree visitMatchCase(SLangParser.MatchCaseContext ctx) {
       return new MatchCaseTreeImpl(meta(ctx), expression, body);
     }
 
+    @Override
+    public Tree visitCatchBlock(SLangParser.CatchBlockContext ctx) {
+      ParameterTree parameter = ctx.formalParameter() == null ? null : (ParameterTree) visit(ctx.formalParameter());
```
pynicolas: We have no test covering the case where `ctx.formalParameter() == null`.

slang-antlr/src/main/java/com/sonarsource/slang/parser/SLangConverter.java
```diff
@@ -262,6 +265,24 @@ public Tree visitMatchCase(SLangParser.MatchCaseContext ctx) {
...
+    }
+
+    @Override
+    public Tree visitTryExpression(SLangParser.TryExpressionContext ctx) {
+      Tree tryBlock = visit(ctx.block());
+      List<CatchTree> catchTreeList = new ArrayList<>();
+      for (SLangParser.CatchBlockContext catchBlockContext : ctx.catchBlock()) {
+        catchTreeList.add((CatchTree) visit(catchBlockContext));
+      }
+      Tree finallyBlock = ctx.finallyBlock() == null ? null : visit(ctx.finallyBlock());
```
pynicolas: We have no test covering the case where `ctx. finallyBlock() == null`.

slang-api/src/main/java/com/sonarsource/slang/api/CatchTree.java
```diff
@@ -0,0 +1,31 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.sonarsource.slang.api;
+
+import javax.annotation.CheckForNull;
+
+public interface CatchTree extends Tree {
+
+  @CheckForNull
+  ParameterTree catchParameter();
```
pynicolas: Shouldn't we declare that as a `Tree`? For example, it seems that in Ruby, we may have a type and no "parameter" name, e.g.:
```
begin  
  # -  
rescue OneTypeOfException  
  # -  
rescue AnotherTypeOfException  
  # -  
else  
  # No exceptions  
end  
```

PR #90, opened 2018-06-18
Rename plugin to SlangPlugin and settings category to "Kotlin"

pynicolas:


sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinLanguage.java
```diff
@@ -28,15 +28,15 @@
   private Configuration configuration;
 
   public KotlinLanguage(Configuration configuration) {
-    super(KotlinPlugin.LANGUAGE_KEY, KotlinPlugin.LANGUAGE_NAME);
+    super(SlangPlugin.LANGUAGE_KEY, SlangPlugin.LANGUAGE_NAME);
```
christophe-zurn-sonarsource: Should `LANGUAGE_KEY`, `LANGUAGE_KEY` and `REPOSITORY_KEY` be also renamed with a `KOTLIN_` prefix now ? Otherwise it might be confusing to have a `LANGUAGE_KEY` constant in a `SlangPlugin` that has "kotlin" value

pynicolas: True. Fixed in f001744.

PR #91, opened 2018-06-18
 SONARSLANG-37 Rule: Lines should not be too long

andrea-guarino-sonarsource:


sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/ChecksVisitor.java
```diff
@@ -94,6 +98,11 @@ public void reportIssue(Tree tree, String message, List<SecondaryLocation> secon
       reportIssue(tree.metaData().textRange(), message, secondaryLocations, gap);
     }
 
+    @Override
+    public String fileContent() {
+      return inputFileContextExtractor.content(currentCtx.inputFile);
```
pynicolas: `InputFileContentExtractor` was useful when we wanted to support older versions of SQ (before SQ 6.2). I don't see why we need it now.

slang-checks/src/main/java/com/sonarsource/slang/checks/TooLongLineCheck.java
```diff
@@ -0,0 +1,65 @@
...
+import com.sonarsource.slang.impl.TextPointerImpl;
+import com.sonarsource.slang.impl.TextRangeImpl;
+import java.text.MessageFormat;
+import java.util.stream.IntStream;
+import org.sonar.check.Rule;
+import org.sonar.check.RuleProperty;
+
+@Rule(key = "S103")
+public class TooLongLineCheck implements SlangCheck {
+  private static final int DEFAULT_MAXIMUM_LINE_LENGTH = 120;
```
pynicolas: There are too many issues in the "ruling" with this setting. We may add some special configuration to the "ruling" to set the threshold higher. Another approach is to set this default value higher (e.g. 200): that way, the rule would be uncontroversial.

PR #93, opened 2018-06-18
SONARSLANG-41 Rule: Useless "if(true) {...}" and "if(false){...}" blocks should be removed

christophe-zurn-sonarsource:


slang-checks/src/main/java/com/sonarsource/slang/checks/IfConditionalAlwaysTrueOrFalseCheck.java
```diff
@@ -0,0 +1,45 @@
...
+
+@Rule(key = "S1145")
+public class IfConditionalAlwaysTrueOrFalseCheck implements SlangCheck {
+  private static final List<String> BOOLEAN_LITERALS = Arrays.asList("true", "false");
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(IfTree.class, (ctx, ifTree) -> {
+      Tree condition = ifTree.condition();
+      if (condition instanceof LiteralTree && BOOLEAN_LITERALS.contains(((LiteralTree) condition).value())) {
```
andrea-guarino-sonarsource: I think we should cover also the cases
- `if (!true)` and `if(!false)`
- And expression that contains a `false` literal
- Or expression that contains a `true` literal

slang-checks/src/main/java/com/sonarsource/slang/checks/IfConditionalAlwaysTrueOrFalseCheck.java
```diff
@@ -0,0 +1,45 @@
...
+        ctx.reportIssue(condition, "Remove this useless \"if\" statement.");
```
andrea-guarino-sonarsource: Minor comment: maybe we could use the `ifTree.keyword()`  (when it will be merged in master)

christophe-zurn-sonarsource: Good point. I created a Jira ticket for this, as we probably have this hardcoded in other rules too: SONARSLANG-44

PR #94, opened 2018-06-19
Add variable declaration to Slang

marcelo-sousa-sonarsource:


slang-antlr/src/test/java/com/sonarsource/slang/antlr/SLangConverterTest.java
```diff
@@ -75,20 +77,67 @@ public void simple_unary_expression() {
...
+  public void variable_declaration() {
+    Tree tree = converter.parse("int var x;").children().get(0);
+    Tree valueTree = converter.parse("int val x;").children().get(0);
+    Tree anotherTree = converter.parse("int var x;").children().get(0);
+    Tree yetAnotherTree = converter.parse("boolean var x;").children().get(0);
+    assertThat(tree).isInstanceOf(VariableDeclarationTree.class);
+    assertThat(anotherTree).isInstanceOf(VariableDeclarationTree.class);
+
+    VariableDeclarationTree varDeclX = (VariableDeclarationTree) tree;
+    VariableDeclarationTree anotherVarDeclX = (VariableDeclarationTree) anotherTree;
```
christophe-zurn-sonarsource: This cast and the previous `assertThat(anotherTree).isInstanceOf(VariableDeclarationTree.class);` are not necessary because of the `isEquivalentTo`  assert you have just after

slang-antlr/src/test/java/com/sonarsource/slang/antlr/SLangConverterTest.java
```diff
@@ -75,20 +77,67 @@ public void simple_unary_expression() {
...
+    assertThat(varDeclX.children()).hasSize(2);
+    assertTree(varDeclX.type()).isIdentifier("int");
+    assertTree(varDeclX.identifier()).isIdentifier("x");
+    assertTree(varDeclX).isEquivalentTo(anotherVarDeclX);
+    assertTree(varDeclX).isNotEquivalentTo(yetAnotherTree);
+    assertTree(varDeclX).isNotEquivalentTo(valueTree);
+  }
+
+  @Test
+  public void variable_declaration_with_initializer() {
```
christophe-zurn-sonarsource: You could add tests of declaration without type and also without both type and initializer (which is theoretically possible), at least to ensure `int var x = 0`, `var x = 0`, `int var x` and `var x` are not equivalent

slang-antlr/src/main/antlr4/com/sonarsource/slang/parser/SLang.g4
```diff
@@ -55,15 +55,34 @@ variableDeclaratorId
 
 methodBody
   : block 
-  | SEMICOLON 
+  | SEMICOLON
   ;
 
 block
-  :  LCURLY (statementOrExpression semi)* (statementOrExpression semi?)? RCURLY 
+  :  LCURLY (statement semi)* (statement semi?)? RCURLY
```
christophe-zurn-sonarsource: Should we keep `semi` and `semi?` here since now we enforce `;` on top level satements?
I.e. should we enforce semicolon here too?

slang-api/src/main/java/com/sonarsource/slang/impl/VariableDeclarationTreeImpl.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.List;
+
+public class VariableDeclarationTreeImpl extends BaseTreeImpl implements VariableDeclarationTree {
+
+  private final IdentifierTree identifier;
+  private final Tree type;
+  private final Tree initializer;
+  private final boolean isVal;
+
+  public VariableDeclarationTreeImpl(TreeMetaData metaData, IdentifierTree identifier, @CheckForNull Tree type, @CheckForNull Tree initializer, boolean isVal) {
```
christophe-zurn-sonarsource: We use `@Nullable` annotation for parameters. `@CheckForNull` has a slightly different meaning

slang-api/src/main/java/com/sonarsource/slang/utils/SyntacticEquivalence.java
```diff
@@ -89,7 +90,8 @@ private static boolean hasDifferentFields(Tree first, Tree second) {
     boolean unaryTreeCheck = (first instanceof UnaryExpressionTree) && ((UnaryExpressionTree) first).operator() != ((UnaryExpressionTree) second).operator();
     boolean binaryTreeCheck = (first instanceof BinaryExpressionTree) && (((BinaryExpressionTree) first).operator() != ((BinaryExpressionTree) second).operator());
     boolean assignTreeCheck = (first instanceof AssignmentExpressionTree) && (((AssignmentExpressionTree) first).operator() != ((AssignmentExpressionTree) second).operator());
-    return nativeTreeCheck || unaryTreeCheck || binaryTreeCheck || assignTreeCheck;
+    boolean vardeclTreeCheck = (first instanceof VariableDeclarationTree) && (((VariableDeclarationTree) first).isVal() != ((VariableDeclarationTree) second).isVal());
```
christophe-zurn-sonarsource: Unit tests for SyntacticEquivalence should be updated too

slang-api/src/test/java/com/sonarsource/slang/impl/VariableDeclarationTreeImplTest.java
```diff
@@ -0,0 +1,57 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class VariableDeclarationTreeImplTest {
+
+  private class TypeNativeKind implements NativeKind {}
+
+  @Test
+  public void test() {
+    TreeMetaData meta = null;
+    Tree variableType = new NativeTreeImpl(meta, new VariableDeclarationTreeImplTest.TypeNativeKind(), null);
```
christophe-zurn-sonarsource: why the `VariableDeclarationTreeImplTest.` prefix?

sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -181,11 +184,30 @@ private Tree convertElementToSlangAST(PsiElement element, TreeMetaData metaData)
...
       return createParameter((KtParameter) element);
+    } else if (element instanceof KtProperty) {
+      return createVariableDeclaration((KtProperty) element);
     } else {
       return convertElementToNative(element, metaData);
     }
   }
 
+  private Tree createVariableDeclaration(KtProperty ktProperty) {
+    TreeMetaData metaData = getTreeMetaData(ktProperty);
```
christophe-zurn-sonarsource: you could directly pass the metada as parameter as it is computed in parent method, the way it is done for other `create` methods.

slang-api/src/main/java/com/sonarsource/slang/visitors/TreePrinter.java
```diff
@@ -54,6 +56,8 @@ public static String tree2string(Tree tree) {
         sb.append(" ").append(((IdentifierTree) t).name());
       } else if (t instanceof NativeTree) {
         sb.append(" ").append(((NativeTree) t).nativeKind());
+      } else if (t instanceof VariableDeclarationTree) {
+        sb.append(" ").append(((VariableDeclarationTree) t).identifier().name());
```
christophe-zurn-sonarsource: I don't think it is necessary to append the identifier name here since it will be printed anyway when iterating over the declaration children

sonar-slang-plugin/src/test/java/com/sonarsource/slang/kotlin/KotlinConverterTest.java
```diff
@@ -102,6 +104,35 @@ public void testNullParameterNames() {
...
+
+  @Test
+  public void testVariableDeclarationWithInitializer() {
+    Tree varX = kotlinStatement("var x : Int = 0");
+    Tree valY = kotlinStatement("val x : Int = \"4\"");
+    assertTree(varX).isInstanceOf(VariableDeclarationTree.class);
+    assertTree(valY).isInstanceOf(VariableDeclarationTree.class);
+    assertThat(((VariableDeclarationTree) varX).initializer()).isInstanceOf(LiteralTree.class);
+    assertThat(((VariableDeclarationTree) valY).initializer()).isInstanceOf(StringLiteralTree.class);
+    assertTree(varX).isNotEquivalentTo(valY);
```
christophe-zurn-sonarsource: One `isEquivalentTo` assert would be nice too here, to ensure equivalence works correctly for declarations with initializers

slang-checks/src/test/resources/com/sonarsource/slang/checks/TooLongLine_40.slang
```diff
@@ -1,4 +1,4 @@
 int fun (p1, p2, p3) {}
 
 // Noncompliant@+1
-int funWithLongName(p1, p2, p3) { println(10);}
\ No newline at end of file
+int fun funWithLongName(p1, p2, p3) { println(10);}
```
christophe-zurn-sonarsource: Nice catch :+1:

slang-antlr/src/main/antlr4/com/sonarsource/slang/parser/SLang.g4
```diff
@@ -55,15 +55,34 @@ variableDeclaratorId
...
+  :  simpleType? declarationModifier identifier ('=' expression)?
+  ;
+
+declarationModifier
+  : VAR
+  | VAL
+  ;
+
+assignment
+  :  expression (assignmentOperator statement)+
```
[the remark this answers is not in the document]
marcelo-sousa-sonarsource: I suggest we leave it as such for simplicity and to allow multiple assignments (x = y = z).

slang-antlr/src/main/antlr4/com/sonarsource/slang/parser/SLang.g4
marcelo-sousa-sonarsource: Yes, I think this is good.

slang-antlr/src/main/antlr4/com/sonarsource/slang/parser/SLang.g4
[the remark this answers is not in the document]
christophe-zurn-sonarsource: Ok for me. (I agree that `x = y = z` makes sense, I was talking about `x = int var y = 0`, but as you said, we can keep it for simplicity)

PR #96, opened 2018-06-19
SONARSLANG-40 Rule S1186: Functions should not be empty

pynicolas:


sonar-slang-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S1186.json
```diff
@@ -0,0 +1,16 @@
+{
+  "title": "Methods should not be empty",
```
christophe-zurn-sonarsource: I think we should update this RSPEC title and description (method -> function) to be consistent across rules related to functions in Kotlin

PR #97, opened 2018-06-19
SONARSLANG-45 Map unary expression from Kotlin to Slang AST

christophe-zurn-sonarsource:


christophe-zurn-sonarsource:
@marcelo-sousa-sonarsource Do you agree with the slight change in the grammar?
This is to allow expressions of the form `!!x` in Slang, without needing parentheses

slang-antlr/src/test/java/com/sonarsource/slang/antlr/SLangConverterTest.java
```diff
@@ -64,7 +64,7 @@ public void simple_binary_expression() {
 
   @Test
   public void simple_unary_expression() {
-    BinaryExpressionTree binary = parseBinary("!(!x) && !(y && z);");
+    BinaryExpressionTree binary = parseBinary("!!x && !(y && z)");
```
marcelo-sousa-sonarsource: Add ; to the end of the slang string code.

sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
```diff
@@ -368,16 +384,27 @@ private Tree createBinaryExpression(TreeMetaData metaData, KtBinaryExpression el
...
     } else {
       // FIXME ensure they are all supported. Ex: Add '/=' for assignments
-      return createOperationExpression(metaData, element);
+      return createNativeOperationExpression(metaData, element);
+    }
+  }
+
+  private Tree createUnaryExpression(TreeMetaData metaData, KtUnaryExpression element) {
+    Tree operand = createElement(element.getBaseExpression());
+    IElementType operationToken = element.getOperationToken();
```
marcelo-sousa-sonarsource: Why is the way we get the operationToken different in the unary and binary expressions?

sonar-slang-plugin/src/test/java/com/sonarsource/slang/kotlin/KotlinConverterTest.java
```diff
@@ -84,6 +84,9 @@ public void testUnmappedBinaryExpression() {
 
   @Test
   public void unaryExpressions() {
+    assertTree(kotlinStatement("!x")).isEquivalentTo(kotlinStatement("!x"));
+    assertTree(kotlinStatement("!x")).isEquivalentTo(slangStatement("!x"));
```
marcelo-sousa-sonarsource: Missing ; in the slang code string

sonar-slang-plugin/src/test/java/com/sonarsource/slang/kotlin/KotlinConverterTest.java
```diff
@@ -84,6 +84,9 @@ public void testUnmappedBinaryExpression() {
...
+    assertTree(kotlinStatement("!!x")).isEquivalentTo(slangStatement("!!x"));
```
marcelo-sousa-sonarsource: Missing ; in the slang code string

sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/KotlinTreeVisitor.java
christophe-zurn-sonarsource: For some reason `getOperationToken()` do not return the same object when called on `KtUnaryExpression ` or `KtBinaryExpression`.
Binary espression returns a `KtOperationReferenceExpression`, which has the `getOperationSignTokenType()` method returning a typed `KtToken`.
But unary expression `getOperationToken()` returns a `KtSimpleNameExpression` instead. So I decided to use the `getOperationToken();` method of the expression.
Basically, it is possible to get a `KtToken` type directly from binary expression, but not in unary expression.

marcelo-sousa-sonarsource: Do we want to change the BinaryExpression method to follow the same logic as UnaryExpression for consistency?

christophe-zurn-sonarsource: I'd prefer to keep it as it is as it has the type safety to be a `KtToken` for binary epxression.
I'd have liked the same thing for unary expression, but it is not available for some reason

PR #98, opened 2018-06-19
SONARSLANG-34 CPD

pynicolas:


sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/CommentAndTokenVisitor.java
```diff
@@ -51,6 +56,13 @@
   private static final int DOC_COMMENT_SUFFIX_LENGTH = 2;
   private static final int LINE_COMMENT_PREFIX_LENGTH = 2;
 
+  private static final Set<KtToken> LITERAL_TOKEN_TYPES = new HashSet<>(Arrays.asList(
+    KtTokens.INTEGER_LITERAL,
+    KtTokens.FLOAT_LITERAL,
```
christophe-zurn-sonarsource: Are you sure that `INTEGER_LITERAL` and `FLOAT_LITERAL` should be mapped to `"LITERAL"` too? In other analyzers (Swift / TSQL) I have only seen the strings mapped to `"LITERAL"`

sonar-slang-plugin/src/test/java/com/sonarsource/slang/kotlin/KotlinConverterTest.java
```diff
@@ -322,9 +325,11 @@ public void testAssignments() {
 
   @Test
   public void testTokens() {
-    List<Token> tokens = kotlin("private fun foo() { }").metaData().tokens();
-    assertThat(tokens).extracting(Token::text).containsExactly("private", "fun", "foo", "(", ")", "{", "}");
-    assertThat(tokens).extracting(Token::isKeyword).containsExactly(true, true, false, false, false, false, false);
+    List<Token> tokens = kotlin("private fun foo() { 42 + \"a\" }").metaData().tokens();
+    assertThat(tokens).extracting(Token::text).containsExactly(
+      "private", "fun", "foo", "(", ")", "{", "42", "+", "\"", "a", "\"", "}");
```
christophe-zurn-sonarsource: Not sure about this, but I would have expected to see a single token `"\"a\""` instead of the 3 tokens for the string literal `"a"`, as in the `SlangConverter`.
Although this is not directly related to this PR but tokens

pynicolas: This is due to the fact that the Kotlin parser parses all string literals as templates.

sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/CommentAndTokenVisitor.java
pynicolas: Fixed in 097155a

slang-antlr/src/main/java/com/sonarsource/slang/parser/SLangConverter.java
```diff
@@ -87,13 +87,6 @@
     SLangParser.THIS
   ));
 
-  private static final Set<Integer> LITERAL_TOKEN_TYPES = new HashSet<>(Arrays.asList(
-    SLangParser.IntegerLiteral,
-    SLangParser.BooleanLiteral,
-    SLangParser.CharacterLiteral,
```
christophe-zurn-sonarsource: Other analyzers also consider character literal as strings for CPD. Up to you if you want to include it or not, since strictly the documentation only mention strings...

sonar-slang-plugin/src/main/java/com/sonarsource/slang/kotlin/CommentAndTokenVisitor.java
```diff
@@ -81,8 +70,8 @@ public void visitElement(@NotNull PsiElement element) {
       Token.Type type = Token.Type.OTHER;
       if (leaf.getElementType() instanceof KtKeywordToken) {
         type = Token.Type.KEYWORD;
-      } else if (LITERAL_TOKEN_TYPES.contains(leaf.getElementType())) {
-        type = Token.Type.LITERAL;
+      } else if (leaf.getElementType() == KtTokens.REGULAR_STRING_PART) {
```
christophe-zurn-sonarsource: If you decide to add character literal, you can use `KtTokens.STRINGS.contains()` here

slang-antlr/src/main/java/com/sonarsource/slang/parser/SLangConverter.java
pynicolas: Let's not consider character literals as strings. I don't think it would have real value.

PR #101, opened 2018-06-21
SONARSLANG-42 Rule: "switch" statements should not have too many "cas…

pynicolas:
…e" clauses


sonar-slang-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S1479.html
```diff
@@ -0,0 +1,3 @@
+<p>When <code>when</code> statements have a large number of clauses, it is usually an attempt to map two sets of data. A real map structure would be
```
andrea-guarino-sonarsource: What do you think about adding a code example?
Btw, I'm not sure this sentence is really suitable for Kotlin. `When` statements seems to be more powerful than standard `switch` (like pattern matching on a given type wouldn't be replace by a map structure).

PR #102, opened 2018-06-21
SONARSLANG-52 Add classes to slang

marcelo-sousa-sonarsource:


slang-api/src/main/java/com/sonarsource/slang/api/ClassDeclarationTree.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.sonarsource.slang.api;
+
+public interface ClassDeclarationTree extends Tree {
+
+  Tree classTree();
```
christophe-zurn-sonarsource: I would have liked to have the `IdentifierTree` in the `ClassDeclarationTree ` too. Otherwise there is no way to retrieve the name of the class easily. WDYT ?

marcelo-sousa-sonarsource: For now, we don't need the name of the class as we only need classes for metrics; adding names will add extra complexity to the Kotlin converter which I would like to avoid for now.

christophe-zurn-sonarsource: Ok. We'll need it at some point for the rule about class name, but we can update API at that time

PR #103, opened 2018-06-21
Fix performance issue with tree tokens and comments

pynicolas:


slang-api/src/main/java/com/sonarsource/slang/impl/TreeMetaDataProvider.java
```diff
@@ -62,18 +69,29 @@ public TextRange textRange() {
...
     @Override
     public List<Token> tokens() {
-      // TODO improve performance by storing an ordered list of tokens
-      return tokens.stream()
-        .filter(token -> token.textRange().isInside(textRange))
-        .collect(Collectors.toList());
+      return getElementsInsideRange(sortedTokens);
+    }
+
+    private <T extends HasTextRange> List<T> getElementsInsideRange(List<T> sortedList) {
```
andrea-guarino-sonarsource: Are we really gaining a lot in term of performance here? If I'm not wrong, worst case analysis would be `log(n) + n` time against `2n` time of the previous implementation.  IMO we're losing a bit in term of readability, not having a huge gain in term of performance. What do you think?

PR #114, opened 2018-06-25
SONARSLANG-49 Rule: IP addresses should not be hardcoded

andrea-guarino-sonarsource:


slang-checks/src/main/java/com/sonarsource/slang/checks/HardcodedIpCheck.java
```diff
@@ -0,0 +1,58 @@
...
+import java.text.MessageFormat;
+import java.util.Arrays;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.sonar.check.Rule;
+
+@Rule(key = "S1313")
+public class HardcodedIpCheck implements SlangCheck {
+
+  private static final String IP_ADDRESS_REGEX = "([^\\d.]*\\/)?(?<ip>(?:\\d{1,3}\\.){3}\\d{1,3}(?!\\d|\\.))(\\/.*)?";
```
pynicolas: Why `\\/` rather than just `/`? (2 times)

andrea-guarino-sonarsource: Fixed in 6ec23101dba28af7b0a02c4842ec0f289de8a10c

PR #115, opened 2018-06-26
SONARSLANG-20 Add cyclomatic complexity metric

andrea-guarino-sonarsource:


sonar-slang-plugin/src/main/java/com/sonarsource/slang/plugin/CyclomaticComplexityVisitor.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public CyclomaticComplexityVisitor() {
+
+    register(FunctionDeclarationTree.class, (ctx, tree) -> {
+      if (tree.name() != null && tree.body() != null) {
+        complexityTrees.add(tree);
+      }
+    });
+
+    register(ClassDeclarationTree.class, (ctx, tree) -> complexityTrees.add(tree));
```
pynicolas: I don't think we should increment on class declarations. Sorry, I missed that one on the first review.

andrea-guarino-sonarsource: no problem. I will remove it

PR #116, opened 2018-06-26
SONARSLANG-56 Rule: Comments should not be empty

pynicolas:


slang-checks/src/test/resources/com/sonarsource/slang/checks/EmptyComment.slang
```diff
@@ -0,0 +1,14 @@
+/* Some comment */
+
+/**/ // Noncompliant
```
christophe-zurn-sonarsource: It would be nice to add a text range check for one of these tests

PR #118, opened 2018-06-26
 Rule: Control-flow statements not nested deeply

marcelo-sousa-sonarsource:


slang-checks/src/main/java/com/sonarsource/slang/checks/TooDeeplyNestedStatementsCheck.java
```diff
@@ -0,0 +1,123 @@
...
+  private static Token getNodeToHighlight(Tree tree) {
+    if (tree instanceof IfTree) {
+      return ((IfTree) tree).ifKeyword();
+    } else if (tree instanceof MatchTree) {
+      return ((MatchTree) tree).keyword();
+    } else if (tree instanceof ExceptionHandlingTree) {
+      return ((ExceptionHandlingTree) tree).tryKeyword();
+    } else if (tree instanceof LoopTree) {
+      return ((LoopTree) tree).keyword();
+    } else {
```
christophe-zurn-sonarsource: I don't think it can ever happen to reach this clause given the logic in the code

PR #121, opened 2018-06-27
SONARSLANG-60 Rule: Unused "private" methods should be removed

christophe-zurn-sonarsource:


slang-checks/src/main/java/com/sonarsource/slang/checks/UnusedPrivateMethodCheck.java
```diff
@@ -0,0 +1,79 @@
...
+        .collect(Collectors.toSet());
+      Set<IdentifierTree> usedIdentifiers = classDeclarationTree.descendants()
+        .filter(IdentifierTree.class::isInstance)
+        .map(IdentifierTree.class::cast)
+        .collect(Collectors.toSet());
+
+      usedIdentifiers.removeAll(classMethods.stream()
+        .map(FunctionDeclarationTree::name)
+        .collect(Collectors.toSet()));
+      usedIdentifiers.remove(classDeclarationTree.identifier());
```
pynicolas: Why? Removing this class doesn't make unit tests fail.

its/ruling/src/test/resources/expected/kotlin-S1144.json
```diff
@@ -0,0 +1,82 @@
+{
+'kotlin-project:corda/core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt':[
+42,
```
pynicolas: That looks a lot like a FP. A safer approach would be to work at file level rather than at class level: for each private method in a file, check whether its name is used somewhere else in the same file.

christophe-zurn-sonarsource: Actually, here the problem comes from the fact that there are missing elements in this AST.
This stems from: `class A(a: int) { constructor(): this(getValue(0)) {} }` -> the delegation of the secondary constructor to the primary constructor (`: this(getValue(0))`) is missing from AST, so this FP arises from the fact that the use of the private function is inside the delegation call.
It is the same for delegation to superclass constructor when there is no primary constructor `class B: A { constructor(a: Int): super(getValue(a)) }`

slang-checks/src/main/java/com/sonarsource/slang/checks/UnusedPrivateMethodCheck.java
christophe-zurn-sonarsource: True, this was related to constructor call method in Kotlin actually, but it is not needed anymore, as 
1. It could not be tested in slang
2. We'll update API to consider constructor as native elements instead of `FunctionDeclarationTree` as discussed

its/ruling/src/test/resources/expected/kotlin-S1144.json
christophe-zurn-sonarsource: This FP will be fixed when fixing the AST creation (https://jira.sonarsource.com/browse/SONARSLANG-70)

PR #124, opened 2018-06-29
SONARSLANG-70 Fix missing element in AST

christophe-zurn-sonarsource:


slang-api/src/main/java/com/sonarsource/slang/api/FunctionDeclarationTree.java
```diff
@@ -32,11 +32,14 @@
...
   IdentifierTree name();
 
-  List<ParameterTree> formalParameters();
+  List<Tree> formalParameters();
 
   @CheckForNull
   BlockTree body();
 
+  @CheckForNull
+  Tree nativeChildren();
```
pynicolas: Why not a `List<Tree>`? Otherwise, it should be named `nativeChild`.

PR #126, opened 2018-07-02
SONARSLANG-18 Compute "cognitive_complexity" metric (Cognitive Complexity)

christophe-zurn-sonarsource:


christophe-zurn-sonarsource:
This branch should be rebased & merged instead of squashed as the first commit is relative to a fix/improvement of cognitive complexity computation

PR #131, opened 2018-07-03
SONARSLANG-57 FP: "main" function parameters shouldn't be considered as unused

christophe-zurn-sonarsource:


slang-checks/src/main/java/com/sonarsource/slang/checks/UnusedFunctionParameterCheck.java
```diff
@@ -26,20 +26,25 @@
...
 import java.util.stream.Collectors;
+import org.sonar.check.Rule;
 
 import static com.sonarsource.slang.utils.SyntacticEquivalence.areEquivalent;
 
 @Rule(key = "S1172")
 public class UnusedFunctionParameterCheck implements SlangCheck {
 
+  // Currently we ignore all functions named "main", however this should be configurable based on the analyzed language in the future.
+  private static final Pattern IGNORED_PATTERN = Pattern.compile("main", Pattern.CASE_INSENSITIVE);
```
andrea-guarino-sonarsource: I think that at some point we could use the same approach used in CommentedCodeCheck (i.e. passing those language dependent features at construction time). But for the time being I'm fine with hardcoding "main".

PR #133, opened 2018-07-04
SONARSLANG-78 Rule: Expressions should not be too complex

christophe-zurn-sonarsource:


slang-checks/src/main/java/com/sonarsource/slang/checks/TooComplexExpressionCheck.java
```diff
@@ -0,0 +1,91 @@
...
+        return true;
+      }
+    }
+    return true;
+  }
+
+  private static int computeExpressionComplexity(BinaryExpressionTree tree) {
+    List<Tree> complexityTrees = new ArrayList<>();
+    TreeVisitor<TreeContext> binaryExpressionVisitor = new TreeVisitor<>();
+    binaryExpressionVisitor.register(BinaryExpressionTree.class, (ctx, binaryTree) -> {
```
alban-auzeill: we should not visit all children but recursively descend only if BinaryExpressionTree, UnaryExpressionTree, ParenthesizedExpressionTree
because children can contain a lambda with some && and || operator

slang-checks/src/test/java/com/sonarsource/slang/checks/TooComplexExpressionCheckTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.junit.Test;
+
+public class TooComplexExpressionCheckTest {
+
+  @Test
+  public void test() {
+    Verifier.verify("TooComplexExpression.slang", new TooComplexExpressionCheck());
+  }
+
```
alban-auzeill: Add a test with a different "max" threshold

PR #138, opened 2018-07-04
SONARSLANG-86 Update rules impacted by parenthesis expression

christophe-zurn-sonarsource:


slang-checks/src/test/resources/com/sonarsource/slang/checks/IfConditionalAlwaysTrueOrFalse.slang
```diff
@@ -26,6 +29,8 @@ else if (cond1 || cond2 || true) // Noncompliant
   return 1
 else if (cond && cond2 && !true && cond3) { // Noncompliant
   return 1
+} else if (cond && !(cond2 && (!true && cond3))) { // Noncompliant
```
vilchik-elena: can we test precise location?

christophe-zurn-sonarsource: I'll leave it as it is as the rule highlights the whole condition :)

PR #146, opened 2018-07-06
SONARSLANG-75 Rule S121: Control structures should use curly braces

alban-auzeill:


alban-auzeill:
Implementation of this rule is postponed, it will be implemented after [SONARSLANG-92](https://jira.sonarsource.com/browse/SONARSLANG-92)

PR #147, opened 2018-07-09
 SONARSLANG-61 Rule S1763: Jump statements should not be followed by dead code

vilchik-elena:
Rebase when merging


slang-checks/src/main/java/com/sonarsource/slang/checks/CodeAfterJumpCheck.java
```diff
@@ -0,0 +1,49 @@
...
+    init.register(BlockTree.class, (ctx, blockTree) -> checkStatements(ctx, blockTree.statementOrExpressions()));
+  }
+
+  private static void checkStatements(CheckContext ctx, List<Tree> statementsOrExpressions) {
+    if (statementsOrExpressions.size() < 2) {
+      return;
+    }
+    statementsOrExpressions.subList(0, statementsOrExpressions.size() - 1).stream()
+      .filter(tree -> tree instanceof JumpTree || tree instanceof ReturnTree)
+      .forEach(tree -> ctx.reportIssue(tree, String.format(MESSAGE, tree.metaData().tokens().get(0).text())));
```
andrea-guarino-sonarsource: You could use `tree.keyword()` instead of taking the first token

vilchik-elena: @andrea-guarino-sonarsource but `tree` has type `Tree`, there is no `keyword()` method on it

PR #148, opened 2018-07-09
SONARSLANG-72 Rename com.sonarsource -> org.sonarsource

andrea-guarino-sonarsource:


pynicolas:
LGTM. However, 2 directories should be renamed:
- utils/generate-android-lint-rules/src/main/java/com/sonarsource
- utils/generate-detekt-rules/src/main/java/com/sonarsource

PR #159, opened 2018-07-31
SONARSLANG-100 Ruby plugin skeleton

saberduck:
**Please don't squash commits**


sonar-ruby-plugin/src/test/java/org/sonarsource/slang/ruby/RubyRulesDefinitionTest.java
```diff
@@ -21,6 +21,7 @@
 
 import org.junit.Test;
 import org.sonar.api.server.rule.RulesDefinition;
+import org.sonarsource.slang.ruby.plugin.RubyRulesDefinition;
```
vilchik-elena: you should move tests to another package

PR #163, opened 2018-08-06
Enable S1134 and S1135 in SonarRuby and test metrics

vilchik-elena:
Do not squash when merging


sonar-ruby-plugin/src/main/resources/org/sonar/l10n/ruby/rules/ruby/S1135.html
```diff
@@ -0,0 +1,8 @@
+<p><code>TODO</code> tags are commonly used to mark places where some more code is required, but which the developer wants to implement later.</p>
+<p>Sometimes the developer will not have the time or will simply forget to get back to that tag.</p>
+<p>This rule is meant to track those tags and to ensure that they do not go unnoticed.</p>
+<h2>See</h2>
```
christophe-zurn-sonarsource: The noncompliant section from RSPEC is missing here

PR #169, opened 2018-08-09
Exclude AbstractSensorTest from coverage

saberduck:


vilchik-elena:
wow, I would simply do that through SQ UI, but looks like it's more reliable the way you did it

PR #170, opened 2018-08-09
Remove irrelevant files from metasploit framework in ruling

saberduck:


saberduck:
We removed whole metasploit framework

PR #172, opened 2018-08-13
SONARSLANG-131 Enable in Ruby rule S1479

vilchik-elena:
I'm not sure if it's better to squash or keep 2 commits in master


sonar-ruby-plugin/src/main/java/org/sonarsource/ruby/converter/RubyVisitor.java
```diff
@@ -221,13 +259,21 @@ private TreeMetaData metaData(AstNode node) {
...
     TextRange mainKeywordTextRange = node.textRangeForAttribute(attribute);
     if (mainKeywordTextRange != null) {
-      return metaDataProvider.metaData(mainKeywordTextRange).tokens().get(0);
+      return Optional.of(metaDataProvider.metaData(mainKeywordTextRange).tokens().get(0));
+    }
+    return Optional.empty();
+  }
+
+  private Token getTokenByAttribute(AstNode node, String attribute) {
+    Optional<Token> token = lookForTokenByAttribute(node, attribute);
```
saberduck: you can use `Optional.orElseThrow` here, it's bit more concise.

PR #174, opened 2018-08-13
Update tests to fix quality gate

christophe-zurn-sonarsource:


christophe-zurn-sonarsource:
Note that `equals` and `hashCode` methods where removed from classes extending `RubyObject`, as JRuby defines `equals` for `RubyObject` as `==` (identity equality). So it does not make sense to override `equals` method for such objects

PR #177, opened 2018-08-14
Enable in Ruby "if" rules

christophe-zurn-sonarsource:


sonar-ruby-plugin/src/main/java/org/sonarsource/ruby/converter/RubyVisitor.java
```diff
@@ -120,6 +128,61 @@ public IRubyObject onDefs(ThreadContext context, IRubyObject node) {
...
   }
 
+  @JRubyMethod(name = "on_if")
+  public IRubyObject onIf(ThreadContext context, IRubyObject node) {
+    IRubyObject updatedNode = visitNode(context, "on_if", node);
+
+    Token mainKeyword = noLocation(updatedNode, "keyword") ? null : getMetaData(updatedNode, "keyword").tokens().get(0);
+    if (mainKeyword == null || mainKeyword.text().equals("unless")) {
+      // Ternary operator and "unless" are not considered as "IfTree" for now
+      return createNativeTreeAdapter(updatedNode);
```
vilchik-elena: IMO it's cleaner to return `node` (then `process` should do the trick of native creation)

sonar-ruby-plugin/src/main/java/org/sonarsource/ruby/converter/RubyVisitor.java
```diff
@@ -120,6 +128,61 @@ public IRubyObject onDefs(ThreadContext context, IRubyObject node) {
...
+    }
+
+    Token elseKeyword = noLocation(updatedNode, "else") ? null : getMetaData(updatedNode, "else").tokens().get(0);
```
vilchik-elena: I'm not sure it's a good idea to have shared logic for first `noLocation` used in `process` and this one. There is only one line common, and semantic is different

slang-checks/src/main/java/org/sonarsource/slang/checks/ElseIfWithoutElseCheck.java
```diff
@@ -19,19 +19,19 @@
...
 import org.sonarsource.slang.checks.api.InitContext;
 import org.sonarsource.slang.checks.api.SlangCheck;
 import org.sonarsource.slang.impl.TextRangeImpl;
-import org.sonar.check.Rule;
 
 @Rule(key = "S126")
 public class ElseIfWithoutElseCheck implements SlangCheck {
 
-  private static final String MESSAGE_TEMPLATE = "Add the missing \"%s\" clause.";
+  private static final String MESSAGE = "Add the missing \"else\" clause.";
```
vilchik-elena: why?

sonar-ruby-plugin/src/main/resources/org/sonar/l10n/ruby/rules/ruby/S1066.html
```diff
@@ -0,0 +1,21 @@
+<p>Merging collapsible <code>if</code> statements increases the code's readability.</p>
```
vilchik-elena: You don't need to put description section to lang subtask if it's the same, it's a separate section and it will be "inherited" during generation

sonar-ruby-plugin/src/main/resources/org/sonar/l10n/ruby/rules/ruby/S1066.html
```diff
@@ -0,0 +1,21 @@
...
+</pre>
+<h2>Compliant Solution</h2>
+<pre>
+def isFileOrDirectory(filename)
+  File.file?(filename) || File.directory?(filename)
+end
+# ...
+
+if !filename.nil? &amp;&amp; isFileOrDirectory(filename)
+    # ...
```
vilchik-elena: why 4 spaces indent?

slang-checks/src/main/java/org/sonarsource/slang/checks/CollapsibleIfStatementsCheck.java
```diff
@@ -31,7 +31,7 @@
 
 @Rule(key = "S1066")
 public class CollapsibleIfStatementsCheck implements SlangCheck {
-  private static final String MESSAGE_TEMPLATE = "Merge this %s statement with the nested one.";
+  private static final String MESSAGE_TEMPLATE = "Merge this \"%s\" statement with the nested one.";
```
vilchik-elena: I did the same for some rule and used single quotes. We should have some style here:)

slang-checks/src/main/java/org/sonarsource/slang/checks/ElseIfWithoutElseCheck.java
christophe-zurn-sonarsource: As discussed yesterday, ruby has an "elsif" keyword, so using this keyword for the message would be confusing, as the rule is about the final "else" clause

sonar-ruby-plugin/src/main/resources/org/sonar/l10n/ruby/rules/ruby/S1066.html
christophe-zurn-sonarsource: Yep, it was a copy-paste mistake. Removed description section from subtask

sonar-ruby-plugin/src/main/resources/org/sonar/l10n/ruby/rules/ruby/S1066.html
christophe-zurn-sonarsource: Ooops :) Fixed it

slang-checks/src/main/java/org/sonarsource/slang/checks/CollapsibleIfStatementsCheck.java
christophe-zurn-sonarsource: Oh ok. I used double quote as the message for the other rule was also using double quotes.
I agree we should be consistent across messages... I created a [ticket](https://jira.sonarsource.com/browse/SONARSLANG-149) to harmonize this.

sonar-ruby-plugin/src/main/java/org/sonarsource/ruby/converter/RubyVisitor.java
```diff
@@ -143,7 +198,7 @@ private NativeTree createNativeTree(AstNode node, List<Object> children) {
     return new NativeTreeImpl(metaData(node), new RubyNativeKind(node.type()), nonNullChildren);
   }
 
-  private static Stream<Tree> treeForChild(@Nullable Object child, TreeMetaData treeMetaData) {
```
saberduck: why `@Nullable` is removed? I think it should be there

christophe-zurn-sonarsource: Oops. Mistake when rebasing on master. Fixed it :+1:

PR #178, opened 2018-08-14
Expressions

vilchik-elena:


sonar-ruby-plugin/src/main/java/org/sonarsource/ruby/converter/RubyVisitor.java
```diff
@@ -49,6 +59,26 @@
 
   private final TreeMetaDataProvider metaDataProvider;
 
+  private static final Map<String, Operator> BINARY_OPERATOR_MAP = Collections.unmodifiableMap(Stream.of(
```
saberduck: This is bit unusual way to build a map, I suggest we either add dependency on Guava and use ImmutableMap.of , or use HashMap put . I think adding guava is fine

sonar-ruby-plugin/src/main/java/org/sonarsource/ruby/converter/adapter/SourceMapAdapter.java
```diff
@@ -36,4 +37,10 @@ public RangeAdapter getRange(String attribute) {
     return new RangeAdapter(runtime, getFromUnderlying(attribute, IRubyObject.class));
   }
 
+  public boolean isPresentFromUnderlying(String attribute) {
```
saberduck: this seems to be unused now

sonar-ruby-plugin/src/test/java/org/sonarsource/ruby/converter/visitor/BinaryExpressionVisitorTest.java
```diff
@@ -0,0 +1,63 @@
...
+    assertTree(rubyStatement("a - b")).isBinaryExpression(Operator.MINUS);
+    assertTree(rubyStatement("a * b")).isBinaryExpression(Operator.TIMES);
+    assertTree(rubyStatement("a / b")).isBinaryExpression(Operator.DIVIDED_BY);
+  }
+
+  @Test
+  public void logical() {
+    assertTree(rubyStatement("a && b")).isBinaryExpression(Operator.CONDITIONAL_AND);
+    assertTree(rubyStatement("a || b")).isBinaryExpression(Operator.CONDITIONAL_OR);
+    // fixme: pairs &&/and and ||/or don't have the same priority, so I'm not sure we should create the same tree
```
saberduck: should we have a ticket for this issue? I am bit worried that this fixme will just stay here forever. Let's either have a ticket, or drop a fixme

PR #183, opened 2018-08-16
SONARSLANG-138 Enable in Ruby rule S1192

saberduck:


sonar-ruby-plugin/src/main/java/org/sonarsource/ruby/converter/RubyVisitor.java
```diff
@@ -91,11 +94,23 @@
...
   public RubyVisitor(TreeMetaDataProvider metaDataProvider) {
     this.metaDataProvider = metaDataProvider;
   }
 
-  public Tree visitNode(AstNode node, List<Object> children) {
+  public void beforeVisit(AstNode node) {
+    nodeTypeStack.push(node.type());
+  }
+
+  public Tree afterVisit(AstNode node, List<Object> children) {
```
christophe-zurn-sonarsource: I find it strange to have the `afterVisit` method calling `visitNode`.
I'd prefer to have all 3 methods called individually by the ruby processor one after the other

sonar-ruby-plugin/src/test/java/org/sonarsource/ruby/converter/visitor/StringLiteralVisitorTest.java
```diff
@@ -0,0 +1,60 @@
...
+      "        redirect asset_path('#{asset}', protocol: 'http')\n" +
+      "      end\n" +
+      "    CODE\n");
+    assertTree(tree).hasChildren(7);
+    Class[] childrenClasses = new Class[7];
+    Arrays.fill(childrenClasses, NativeTree.class);
+    assertTree(tree).hasChildren(childrenClasses);
+  }
+
+
```
christophe-zurn-sonarsource: We could add a test for interpolated strings such as `"foo#{bar}baz"`

sonar-ruby-plugin/src/test/java/org/sonarsource/ruby/converter/AbstractRubyConverterTest.java
```diff
@@ -81,6 +83,14 @@ protected static LiteralTree literal(String value) {
     return new LiteralTreeImpl(null, value);
   }
 
+  protected static StringLiteralTree stringLiteral(String value) {
```
christophe-zurn-sonarsource: I think you can remove this method as it is not used anywhere anymore

sonar-ruby-plugin/src/main/resources/whitequark_parser_init.rb
```diff
@@ -47,8 +47,9 @@ def initialize(visitor)
 
   def process(node)
     return if node.nil?
+    @visitor.beforeVisit(AstNode.new(node))
     node = super
-    @visitor.visitNode(AstNode.new(node), node.to_a)
+    @visitor.afterVisit(AstNode.new(node), node.to_a)
```
christophe-zurn-sonarsource: See my other comment about calling visit methods individually.
But I think I'd prefer to have something like 
```
tree = @visitor.visitNode(AstNode.new(node), node.to_a)
@visitor.afterVisit(tree)
tree
```

sonar-ruby-plugin/src/test/java/org/sonarsource/ruby/converter/visitor/StringLiteralVisitorTest.java
```diff
@@ -0,0 +1,60 @@
...
+  @Test
+  public void plain_string_literal() {
+    StringLiteralTree tree = (StringLiteralTree) rubyStatement("'foo'");
+    assertTree(tree).isStringLiteral("foo");
+  }
+
+  @Test
+  public void file_macro() {
+    NativeTree tree = (NativeTree) rubyStatement("__FILE__");
+    assertThat(((NativeTree) tree.children().get(0)).nativeKind()).isEqualTo(nativeKind("(SonarRuby analysis)"));
```
christophe-zurn-sonarsource: Is it ok to leave native kind as `"(SonarRuby analysis)"` ? we could maybe set it to something more explicit such as `"FILE_MACRO"`.
Not sure what's best... you can leave it as is if you prefer it like that :)

saberduck: I would keep this, should be obvious what this means

christophe-zurn-sonarsource: Ok. Fine by me :)

PR #184, opened 2018-08-16
SONARSLANG-137 Enable in Ruby rule S1313

saberduck:


sonar-ruby-plugin/src/main/resources/org/sonar/l10n/ruby/rules/ruby/S1313.html
```diff
@@ -0,0 +1,31 @@
...
+  <li> it places the responsibility of setting the value to use in production on the shoulders of the developer </li>
+  <li> it allows attackers to decompile the code and thereby discover a potentially sensitive address </li>
+</ul>
+<h2>Recommended Secure Coding Practices</h2>
+<ul>
+  <li> make the IP address configurable. </li>
+</ul>
+<h2>Noncompliant Code Example</h2>
+<pre>
+String ip = "192.168.12.42"; // Noncompliant
```
christophe-zurn-sonarsource: Noncompliant section of RSPEC should be updated with example for ruby

PR #186, opened 2018-08-17
SONARSLANG-151 Enable in Ruby rule S3776 (FunctionCognitiveComplexityCheck) and Cognitive complexity metric

christophe-zurn-sonarsource:


sonar-ruby-plugin/src/main/java/org/sonarsource/ruby/converter/RubyVisitor.java
```diff
@@ -115,8 +126,9 @@ public Tree visitNode(AstNode node, List<Object> children) {
       case "and":
         return createLogicalOperation(node, children, Operator.CONDITIONAL_AND);
       case "begin":
-      case "kwbegin":
         return createFromBeginNode(node, children);
+      case KW_BEGIN:
```
saberduck: I find it bit strange to introduce a constant only for one node type. Either keep a string, or let's do it for all of them (perhaps we can also have an enum for it)

christophe-zurn-sonarsource: Indeed, it does not make much sense to have a constant for a single case...
I kept it as a string for now.

PR #187, opened 2018-08-18
SONARSLANG-152 Enable in Ruby rule S117

saberduck:


slang-plugin/src/main/java/org/sonarsource/slang/plugin/RulesDefinitionUtils.java
```diff
@@ -0,0 +1,62 @@
...
+import org.sonarsource.slang.checks.utils.Language;
+import org.sonarsource.slang.checks.utils.PropertyDefaultValue;
+import org.sonarsource.slang.checks.utils.PropertyDefaultValues;
+
+public class RulesDefinitionUtils {
+
+  private RulesDefinitionUtils() {
+  }
+
+  public static void setDefaultValuesForParameters(RulesDefinition.NewRepository repository, List<Class> checks, Language language) {
```
christophe-zurn-sonarsource: This method has no test at the moment

saberduck: It is tested indirectly via Ruby and Kotlin rule definitions, but OK I will added unit test just for this one

sonar-ruby-plugin/src/main/java/org/sonarsource/ruby/converter/RubyVisitor.java
```diff
@@ -114,6 +115,9 @@ public Tree visitNode(AstNode node, List<Object> children) {
     switch (node.type()) {
       case "and":
         return createLogicalOperation(node, children, Operator.CONDITIONAL_AND);
+      case "arg":
+      case "optarg":
```
christophe-zurn-sonarsource: I think `optarg` should wrap the `ParameterTree` around a native tree, otherwise we're loosing information when converting to slang AST (we're losing the parameter default value node)

christophe-zurn-sonarsource:
LGTM

PR #189, opened 2018-08-20
SONARSLANG-108 Rule S1314: Octal values should not be used

christophe-zurn-sonarsource:


slang-api/src/main/java/org/sonarsource/slang/impl/IntegerLiteralTreeImpl.java
```diff
@@ -0,0 +1,58 @@
...
+import org.sonarsource.slang.api.TreeMetaData;
+
+public class IntegerLiteralTreeImpl extends LiteralTreeImpl implements IntegerLiteralTree {
+
+  public IntegerLiteralTreeImpl(TreeMetaData metaData, String value) {
+    super(metaData, value);
+  }
+
+  @Override
+  public boolean isOctal() {
```
saberduck: I think that interpretation of the value is ultimately language specific and should be responsibility of AST converter specific for the language. SLang tree should be created by explicitly saying what kind of integer literal is it.

slang-checks/src/main/java/org/sonarsource/slang/checks/CheckList.java
```diff
@@ -43,7 +43,9 @@
     VariableAndParameterNameCheck.class
   };
 
-  private static final Class[] KOTLIN_CHECK_BLACK_LIST = {};
+  private static final Class[] KOTLIN_CHECK_BLACK_LIST = {
+    OctalValuesCheck.class
```
saberduck: Please create a ticket to enable this rule in Kotlin, so we don't forget about it

its/ruling/src/test/resources/expected/ruby/ruby-S1314.json
```diff
@@ -0,0 +1,180 @@
...
+],
+'ruby-project:sources/ruby/vagrant/test/unit/plugins/hosts/linux/cap/ssh_test.rb':[
+12,
+],
+'ruby-project:sources/ruby/vagrant/test/unit/vagrant/util/ssh_test.rb':[
+15,
+24,
+27,
+],
+'ruby-project:sources/ruby/vagrant/test/unit/vagrant/util/which_test.rb':[
```
saberduck: Looking at results here, I would add two exceptions to the rule - when value is in range 1 - 7, because the result is same regardless the base. And for values related to unix permissions 0000 - 0777 , this is often used and completely expected imo

slang-checks/src/main/java/org/sonarsource/slang/checks/CheckList.java
pynicolas: Kotlin [does not support](https://kotlinlang.org/docs/reference/basic-types.html) octal literals:
> NOTE: Octal literals are not supported.

saberduck: OK, I was not aware, then maybe we can put a comment here?

PR #198, opened 2018-08-22
SONARSLANG-99 SonarRuby supports import of issues raised by Rubocop

alban-auzeill:


sonar-ruby-plugin/src/main/java/org/sonarsource/ruby/externalreport/rubocop/RuboCopSensor.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  private static void saveIssue(SensorContext context, RuboCopJsonReportReader.Issue issue) {
+    if (isEmpty(issue.ruleKey) || isEmpty(issue.filePath) || isEmpty(issue.message)) {
+      LOG.debug("Missing information or unsupported file type for ruleKey:'{}', filePath:'{}', message:'{}'",
+        issue.ruleKey, issue.filePath, issue.message);
+      return;
+    }
+    FilePredicates predicates = context.fileSystem().predicates();
+    InputFile inputFile = context.fileSystem().inputFile(predicates.or(
```
alban-auzeill: Try to replace by predicates.hasPath(...)

its/plugin/projects/externalreport/rubocop/yaml-issue.rb
```diff
@@ -0,0 +1,3 @@
+def security
+  YAML.load('--- foo')
+end
```
alban-auzeill: move this file in a sub-folder, so the paths will contain a path separator and will be tested on windows

sonar-ruby-plugin/pom.xml
```diff
@@ -63,6 +63,10 @@
             <type>gem</type>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
```
alban-auzeill: Try to replace with simple json

sonar-ruby-plugin/src/main/java/org/sonarsource/ruby/externalreport/rubocop/RuboCopSensor.java
```diff
@@ -0,0 +1,120 @@
...
+      .severity(externalRuleLoader.ruleSeverity(ruleKey))
+      .remediationEffortMinutes(externalRuleLoader.ruleConstantDebtMinutes(ruleKey));
+
+    NewIssueLocation primaryLocation = newExternalIssue.newLocation()
+      .message(issue.message)
+      .on(inputFile);
+
+    if (issue.startLine !=null) {
+      boolean rangeIsProvided = issue.startColumn != null && issue.lastLine != null && issue.lastColumn != null;
+      boolean rangeIsValid = rangeIsProvided && (issue.startLine < issue.lastLine || (issue.startLine == issue.lastLine && issue.startColumn <= issue.lastColumn));
```
pynicolas: The range could also be invalid if the file does not contain enough lines or if a line doesn't have enough columns. What do we do in such cases?

sonar-ruby-plugin/src/test/java/org/sonarsource/ruby/externalreport/rubocop/RuboCopSensorTest.java
```diff
@@ -0,0 +1,242 @@
...
+    assertThat(firstRange.end().lineOffset()).isEqualTo(7);
+
+    ExternalIssue second = externalIssues.get(1);
+    assertThat(second.primaryLocation().inputComponent().key()).isEqualTo("rubocop-project:yaml-issue.rb");
+    assertThat(second.ruleKey().toString()).isEqualTo("rubocop:Security/YAMLLoad");
+    assertThat(second.type()).isEqualTo(RuleType.VULNERABILITY);
+    assertThat(second.severity()).isEqualTo(Severity.MAJOR);
+    assertThat(second.primaryLocation().message()).isEqualTo("Security/YAMLLoad: Prefer using `YAML.safe_load` over `YAML.load`.");
+    TextRange secondRange = second.primaryLocation().textRange();
+    assertThat(secondRange.start().line()).isEqualTo(2);
```
pynicolas: The assertions on text ranges would be more readable if we able to write something like 
```
assertThat(secondRange).isEqualTo(2, 7, 2, 11);
```

utils/generate-rubocop-rules/src/main/java/org/sonarsource/slang/externalreport/rubocop/RuboCopRuleDefinition.java
```diff
@@ -0,0 +1,138 @@
...
+
+  private static final Path RUBOCOP_HELP = Paths.get("utils", "generate-rubocop-rules",
+    "src", "main", "resources", "rubocop.yml");
+
+  public static void main(String[] args) throws IOException {
+    String rules = RuboCopRuleDefinitionGenerator.generateRuleDefinitionJson(resolve(RUBOCOP_HELP));
+    Files.write(resolve(RULES_FILE), rules.getBytes(UTF_8));
+  }
+
+  private static Path resolve(Path path) throws IOException {
```
pynicolas: Why do we need that?

utils/generate-rubocop-rules/src/main/java/org/sonarsource/slang/externalreport/rubocop/RuboCopRuleDefinition.java
```diff
@@ -0,0 +1,138 @@
...
+  private static class ExternalRule {
+    String key;
+    String name;
+    String type;
+    String severity;
+    String description;
+    String url;
+  }
+
+  private static class RuboCopHelp {
```
pynicolas: Why "help"?

utils/generate-rubocop-rules/src/main/java/org/sonarsource/slang/externalreport/rubocop/RuboCopRuleDefinition.java
```diff
@@ -0,0 +1,138 @@
...
+      ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+      mapper.readTree(ruboCopHelpPath.toFile()).fields().forEachRemaining(this::addRule);
+    }
+
+    private void addRule(Map.Entry<String, JsonNode> entry) {
+      ExternalRule rule = new ExternalRule();
+      rule.key = entry.getKey();
+      String category = ruleCategory(rule.key);
+      String pascalCaseName = rule.key.substring(category.length() + 1).replaceAll("([^A-Z])([A-Z])", "$1 $2");
+      rule.name = pascalCaseName + " (" + category + ")";
```
pynicolas: We could think of transforming this "category" to a tag. Just an idea.

alban-auzeill: It could be an improvement ... once SQ platform support tags on external issue. Currently we can provide tags information on external issues, but we can not see them by tags in SQ UI.

utils/generate-rubocop-rules/src/main/java/org/sonarsource/slang/externalreport/rubocop/RuboCopRuleDefinition.java
alban-auzeill: The name is related to the `rubocop --show-cops` command line output. I called it help, because it's the rubocop command line help. But in fact its a yaml structured list of rules. So I will change it to `RuboCopRulesYamlFile`

utils/generate-rubocop-rules/src/main/java/org/sonarsource/slang/externalreport/rubocop/RuboCopRuleDefinition.java
alban-auzeill: It's mainly to resolve "sonar-ruby-plugin" folder, without forcing you to execute `mvn exec:java` in the root directory of slang repo.

sonar-ruby-plugin/src/test/java/org/sonarsource/ruby/externalreport/rubocop/RuboCopSensorTest.java
alban-auzeill: changed to the following format `from line 2 offset 7 to line 2 offset 11`

sonar-ruby-plugin/src/main/java/org/sonarsource/ruby/externalreport/rubocop/RuboCopSensor.java
alban-auzeill: `issues_when_rubocop_file_line_error` tests that we catch the exception: `100 is not a valid line for pointer. File useless-assignment.rb has 5 line(s)`. In this case we fail to import all the RuboCop report, but we continue the analysis. And it should never append.

PR #204, opened 2018-08-23
SONARSLANG-173 Shouldn't print ERROR level log for empty file

alban-auzeill:


slang-plugin/src/main/java/org/sonarsource/slang/plugin/SlangSensor.java
```diff
@@ -47,6 +48,7 @@
 
 public abstract class SlangSensor implements Sensor {
   private static final Logger LOG = Loggers.get(SlangSensor.class);
+  private static final Pattern EMPTY_FILE_CONTENT_PATTERN = Pattern.compile("[ \t\r\n]*+");
```
saberduck: you can use `\s`

PR #209, opened 2018-08-24
Fix language casing in Sonarpedia

erichirlemann:


pynicolas:
@erichirlemann Can you please give some context? Why do we want to change anything? Is something broken?

PR #210, opened 2018-08-27
SONARSLANG-176 Reduce size of the Ruby plugin

alban-auzeill:


sonar-ruby-plugin/pom.xml
```diff
@@ -159,7 +167,125 @@
...
+                                        <include>jruby/**</include>
+                                        <include>META-INF/**</include>
+                                        <include>org/jcodings/**</include>
+                                        <include>org/joda/**</include>
+                                        <include>org/joni/**</include>
+                                        <include>org/jruby/**</include>
+                                        <include>tables/CR_Digit.bin</include>
+                                        <include>tables/CR_Space.bin</include>
+                                        <include>tables/Transcoder_Utf1632_ByteArray.bin</include>
+                                        <include>tables/Transcoder_Utf1632_WordArray.bin</include>
```
pynicolas: It seems that we should also include "/tables/CR_Word.bin". I got the following error when analyzing [this file](https://github.com/octobox/octobox/blob/master/lib/search_parser.rb):
```
ERROR: Unable to parse file: file:///Users/pierre-yvesnicolas/test/ruby/octobox/lib/search_parser.rb. Parse error at position 5:24
ERROR: (SyntaxError) entry: /tables/CR_Word.bin not found: /(\w+):[\ \u3000]?([\w\p{Han}\p{Katakana}\p{Hiragana}\p{Hangul}\u30FC\.\-,\/]+|(["'])(\\?.)*?\3)/
```

sonar-ruby-plugin/pom.xml
```diff
@@ -180,8 +306,8 @@
                         <configuration>
                             <rules>
                                 <requireFilesSize>
-                                    <minsize>10000000</minsize>
-                                    <maxsize>30000000</maxsize>
+                                    <minsize>8000000</minsize>
+                                    <maxsize>12000000</maxsize>
```
pynicolas: I would prefer to have a slightly smaller range e.g. 9 to 11M.

PR #211, opened 2018-08-28
 SONARSLANG-97 SonarLint support

saberduck:


sonar-kotlin-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/S3923.html
```diff
@@ -1,5 +1,37 @@
-<p>Having all branches in a <code>when</code> or <code>if</code> chain with the same implementation is an error. Either a copy-paste error was made
-and something different should be executed, or there shouldn't be a <code>when</code>/<code>if</code> chain at all.</p>
+<p>Having all branches in a <code>switch</code> or <code>if</code> chain with the same implementation is an error. Either a copy-paste error was made
```
alban-auzeill: Again this is java **switch** not kotlin **when** :disappointed:

PR #213, opened 2018-08-29
SONARSLANG-95 S1125: Kotlin raises FP for nullable Boolean variables

christophe-zurn-sonarsource:


slang-checks/src/main/java/org/sonarsource/slang/checks/BooleanLiteralCheck.java
```diff
@@ -19,25 +19,23 @@
...
-import org.sonar.check.Rule;
 
 @Rule(key = "S1125")
 public class BooleanLiteralCheck implements SlangCheck {
   private static final List<BinaryExpressionTree.Operator> CONDITIONAL_BINARY_OPERATORS = Arrays.asList(
     BinaryExpressionTree.Operator.CONDITIONAL_AND,
-    BinaryExpressionTree.Operator.CONDITIONAL_OR,
-    BinaryExpressionTree.Operator.EQUAL_TO,
-    BinaryExpressionTree.Operator.NOT_EQUAL_TO);
+    BinaryExpressionTree.Operator.CONDITIONAL_OR);
```
alban-auzeill: IMO we should also temporarily simplify the S1125.html description.

PR #214, opened 2018-08-29
SONARSLANG-178 Create an issue on a file when parsing fails (using S2260)

alban-auzeill:


slang-plugin/src/main/java/org/sonarsource/slang/plugin/SlangSensor.java
```diff
@@ -87,7 +87,7 @@ private static boolean analyseFiles(ASTConverter converter,
         analyseFile(converter, inputFileContext, inputFile, visitors);
       } catch (ParseException e) {
         logParsingError(inputFile, e);
-        inputFileContext.reportError("Unable to parse file: " + inputFile, e.getPosition());
```
saberduck: I think we should call also `newAnalysisError` API for Sonarlint

alban-auzeill: "reportParseError" calls "reportError"
"reportError" calls "newAnalysisError"
so "newAnalysisError" is always called, but to be more explicit on the naming, I renamed:
reportError -> reportAnalysisError
reportParseError -> reportAnalysisParseError

PR #215, opened 2018-08-30
SONARSLANG-149 Slang: fix consistency between issue messages

alban-auzeill:


saberduck:
That's not always true, literal content doesn't contain quotes in Ruby, see here https://peach.sonarsource.com/issues?languages=ruby&open=AWVJjG-P7Xgo2MyxdrOh&resolved=false

PR #217, opened 2018-08-30
SONARSLANG-91 Support throw in S1763

saberduck:


sonar-ruby-plugin/src/test/java/org/sonarsource/ruby/converter/visitor/ExceptionHandlingVisitorTest.java
```diff
@@ -181,4 +181,7 @@ public void only_ensure_block() {
     assertRange(finallyBlock.textRange()).hasRange(3, 4, 3, 5);
   }
 
+
+
+
```
pynicolas: It would be better to exclude that change...

sonar-ruby-plugin/src/test/java/org/sonarsource/ruby/converter/visitor/RaiseVisitorTest.java
```diff
@@ -0,0 +1,48 @@
...
+  public void test_raise() {
+    FunctionDeclarationTree tree = ((FunctionDeclarationTree) rubyStatement("" +
+      "def foo\n" +
+      "  raise 42\n" +
+      "end"));
+    Tree raise = tree.body().statementOrExpressions().get(0);
+    assertTree(raise).isInstanceOf(ThrowTree.class);
+    ThrowTree throwTree = (ThrowTree) raise;
+    assertTree(throwTree.body()).isEquivalentTo(nativeTree(nativeKind("send"), asList(
+      identifier("raise"),
```
pynicolas: Why this identifier?

pynicolas: Let's remove it.

PR #222, opened 2018-09-03
SONARSLANG-181 Fix Ruby AST completeness

alban-auzeill:


alban-auzeill:
Implementation is postponed to next Spring. The test of this branch should pass without a list of exclusions (KNOWN_TYPES_WITH_VALIDATION_ISSUES)

PR #225, opened 2018-09-24
SONARSLANG-182 [Scala] Create sonar-scala-plugin skeleton

alban-auzeill:


sonar-scala-plugin/src/test/java/org/sonarsource/scala/plugin/ScalaLanguageTest.java
```diff
@@ -0,0 +1,41 @@
...
+  public void test_suffixes_default() {
+    ScalaLanguage scalaLanguage = new ScalaLanguage(new MapSettings().asConfig());
+    assertThat(scalaLanguage.getFileSuffixes()).containsExactly(".scala");
+  }
+
+  @Test
+  public void test_suffixes_empty() {
+    ScalaLanguage scalaLanguage = new ScalaLanguage(new MapSettings().setProperty(ScalaPlugin.SCALA_FILE_SUFFIXES_KEY, "").asConfig());
+    assertThat(scalaLanguage.getFileSuffixes()).containsExactly(".scala");
+  }
```
pynicolas: What about testing a non-default setting?

sonar-scala-plugin/src/main/java/org/sonarsource/scala/plugin/ScalaPlugin.java
```diff
@@ -0,0 +1,66 @@
...
+  private static final String GENERAL = "General";
+  private static final String SCALA_CATEGORY = "Scala";
+
+  @Override
+  public void define(Context context) {
+
+    context.addExtensions(
+      ScalaLanguage.class,
+      ScalaSensor.class,
+      ScalaProfileDefinition.class,
```
pynicolas: Maybe we can exclude that for Sonarlint.

sonar-scala-plugin/src/main/resources/org/sonar/l10n/scala/rules/scala/Sonar_way_profile.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "Sonar way",
+  "ruleKeys": [
+    "ParsingError"
```
pynicolas: It means it will be activated in SonarLint: is that what we want?

alban-auzeill: no

PR #233, opened 2018-09-28
SONARSLANG-187 [Scala] CPD

pynicolas:


sonar-scala-plugin/src/main/scala/org/sonarsource/scala/converter/ScalaConverter.scala
```diff
@@ -73,7 +73,21 @@ class ScalaConverter extends slang.api.ASTConverter {
       }
     }
 
-    def convert(trees: scala.List[scala.meta.Tree]): java.util.List[slang.api.Tree] = {
+    private def createTopLevelTree(metaData: TreeMetaData, stats: List[Stat]) = {
+      val firstCpdTokenRange = stats.flatMap(t => if (t.is[scala.meta.Pkg]) t.asInstanceOf[Pkg].stats else List(t))
+        .find(t => t.isNot[scala.meta.Import])
+        .map(t => textRange(t.tokens.apply(0).pos))
```
alban-auzeill: Use Slang AST to find the token

PR #246, opened 2018-10-04
SONARSLANG-219 [Scala] Enable Rule: S1313 Using hardcoded IP addresses is security-sensitive

alban-auzeill:


sonar-scala-plugin/src/main/resources/org/sonar/l10n/scala/rules/scala/S1313.html
```diff
@@ -0,0 +1,39 @@
...
+sensitive address. They can perform a Denial of Service attack on the service at this address or spoof the IP address. Such an attack is always
+possible, but in the case of a hardcoded IP address the fix will be much slower, which will increase an attack's impact.</p>
+<h2>Recommended Secure Coding Practices</h2>
+<ul>
+  <li> make the IP address configurable. </li>
+</ul>
+<h2>Noncompliant Code Example</h2>
+<pre>
+val ip = "192.168.12.42" // Noncompliant
+val socket new Socket(ip, 6667))
```
andrea-guarino-sonarsource: I think you forgot to add `=` and there is an extra parenthesis at the end.

alban-auzeill: Yes of course, the code should be: val socket = new Socket(ip, 6667)

PR #249, opened 2018-10-04
SONARSLANG-241 [Scala] Import coverage from SCoverage

quentin-jaquier-sonarsource:


sonar-scala-plugin/src/main/java/org/sonarsource/scala/plugin/ScoverageSensor.java
```diff
@@ -0,0 +1,182 @@
...
+    List<File> reportFiles = ExternalReportProvider.getReportFiles(context, COVERAGE_REPORT_PATHS_KEY);
+
+    if (reportFiles.isEmpty()) {
+      //Return if we don't have any files
+      return;
+    }
+
+    Set<String> unresolvedInputFile = new HashSet<>();
+
+    for (File f : reportFiles) {
```
alban-auzeill: It's useful for the user to have log like:
```
LOG.info("Importing coverage from {}", f.getPath());
```

sonar-scala-plugin/src/main/java/org/sonarsource/scala/plugin/ScoverageSensor.java
```diff
@@ -0,0 +1,182 @@
...
+        line = Integer.valueOf(attribute.getValue());
+      }
+      if (attribute.getName().equals(SOURCE_ATTRIBUTE)) {
+        source = attribute.getValue();
+      }
+    }
+    if(line != null && invocationCount != null && source != null) {
+      addStatementToMap(linesHitPerFiles, source, line, invocationCount);
+    } else {
+      LOG.warn("Some attributes of statement at line {} of scoverage report are not present.", currentEvent.getLocation().getLineNumber());
```
alban-auzeill: My fear is that: if we import a coverage report with a format that we don't support, if the project has 1M executable line of code, we will have 1M of warning in the log :grimacing:

sonar-scala-plugin/src/test/java/org/sonarsource/scala/plugin/ScoverageSensorTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+    return context;
+  }
+
+  private void createReportFiles(String coverageReportPath, Path baseDir, DefaultFileSystem defaultFileSystem) throws IOException {
+    String[] coverageReportPaths = coverageReportPath.split(",");
+    for (String reportPath : coverageReportPaths) {
+      reportPath = reportPath.trim();
+      // if report is relative path we create it under fake filesystem
+      if (!Paths.get(coverageReportPath).isAbsolute()) {
```
alban-auzeill: Useless test because `coverageReportPath` are always relative in our unit tests. And I would prefer having a test about _file exists_, to remove the following `try/catch`.

sonar-scala-plugin/src/test/java/org/sonarsource/scala/plugin/ScoverageSensorTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+    new ScoverageSensor().execute(context);
+
+    String fileKey1 = MODULE_KEY + ":file1.scala";
+    String fileKey2 = MODULE_KEY + ":file2.scala";
+
+    //File1
+    assertThat(context.lineHits(fileKey1, 6)).isNull();
+    assertThat(context.lineHits(fileKey1, 7)).isEqualTo(1);
+    assertThat(context.lineHits(fileKey1, 8)).isNull();
```
alban-auzeill: only non executable lines are tested with: `.isNull()`
but executable lines, not covered, should be: `.isEqualTo(0)`

PR #250, opened 2018-10-04
SONARSLANG-230 [Scala] Enable S1481 Unused local variables should be removed

pynicolas:


pynicolas:
@andrea-guarino-sonarsource fixed in 44414a0

PR #251, opened 2018-10-04
SONARSLANG-215 [Scala] Enable Rule: S1764 Identical expressions should not be used on both sides of a binary operator

andrea-guarino-sonarsource:


sonar-scala-plugin/src/main/scala/org/sonarsource/scala/converter/ScalaConverter.scala
```diff
@@ -164,6 +185,23 @@ class ScalaConverter extends slang.api.ASTConverter {
...
+      }
+      // Scala has special shorthand for an anonymous function
+      // (_ || _) is equivalent to (p1, p2) => p1 || p2
+      val elems = infix.args :+ infix.lhs
+      if (elems.toStream.exists(term => term.is[Term.Placeholder])) {
+        return createNativeTree(metaData, infix)
+      }
+      val leftOperand = convert(infix.lhs)
+      val rightOperand = convert(infix.args.head)
+      val operatorToken = new TokenImpl(textRange(infix.op), infix.op.value, Token.Type.STRING_LITERAL)
```
pynicolas: I don't really like that. Ideally, we should re-use the token which is already available from the metadata (we may use something similar to the `keyword` function just below). If we recreate a token, we should at least not declare it as `Token.Type.STRING_LITERAL`.

sonar-scala-plugin/src/test/java/org/sonarsource/scala/converter/BinaryExpressionTreeTest.java
```diff
@@ -0,0 +1,94 @@
...
+    assertTree(binaryExpressionTree).isBinaryExpression(BinaryExpressionTree.Operator.NOT_EQUAL_TO);
+
+    binaryExpressionTree = scalaStatement("a > 2");
+    assertTree(binaryExpressionTree).isBinaryExpression(BinaryExpressionTree.Operator.GREATER_THAN);
+
+    binaryExpressionTree = scalaStatement("a >= 2");
+    assertTree(binaryExpressionTree).isBinaryExpression(BinaryExpressionTree.Operator.GREATER_THAN_OR_EQUAL_TO);
+
+    binaryExpressionTree = scalaStatement("a >= 2");
+    assertTree(binaryExpressionTree).isBinaryExpression(BinaryExpressionTree.Operator.GREATER_THAN_OR_EQUAL_TO);
```
pynicolas: These 2 lines are identical to the 2 previous ones.

PR #255, opened 2018-10-05
SONARSLANG-210 [Scala] Enable Rule: S1940 Boolean checks should not be inverted

andrea-guarino-sonarsource:


sonar-scala-plugin/src/main/resources/org/sonar/l10n/scala/rules/scala/S1940.html
```diff
@@ -0,0 +1,12 @@
+<p>It is needlessly complex to invert the result of a boolean comparison. The opposite comparison should be made instead.</p>
+<h2>Noncompliant Code Example</h2>
+<pre>
+if ( !(a == 2)) { ...}  // Noncompliant
+val b = !(i &lt; 10);  // Noncompliant
```
pynicolas: Let's remove the semicolon from both this example and from the compliant solution.

andrea-guarino-sonarsource: Fixed in 23ac2653261412128dca99a855959f7be83eeafd

PR #256, opened 2018-10-05
SONARSLANG-212 [Scala] Enable Rule: S108 Nested blocks of code should not be left empty

andrea-guarino-sonarsource:


andrea-guarino-sonarsource:
> When there's no issue in the ruling, we have no evidence that the rule is working properly for a given language. To address this problem, we add a small file with one issue in the folders of `its/ruling/src/test/resources/sources`.

Fixed in ec258951ae024c9255603932a3005345a3cd94d9

PR #263, opened 2018-10-08
SONARSLANG-218 [Scala] Enable Rule: S1862 Related "if"/"else if" statements and "case" in a "match" should not have the same condition

alban-auzeill:


its/ruling/src/test/resources/expected/scala/scala-S3923.json
```diff
@@ -1,4 +1,7 @@
 {
+'scala-project:ruling/src/test/resources/sources/scala/S1862.scala':[
```
andrea-guarino-sonarsource: Can we avoid this issue?

PR #271, opened 2018-10-09
SONARSLANG-228 [Scala] Enable Rule: S1172 Unused function parameters should be removed

quentin-jaquier-sonarsource:


sonar-scala-plugin/src/test/java/org/sonarsource/scala/converter/FunctionDeclarationTreeTest.java
```diff
@@ -64,4 +65,43 @@ public void function_without_parameter_list() {
   public void function_with_multiple_parameter_lists() {
```
andrea-guarino-sonarsource: Could you add assertion on formal parameters when using multiple parameter lists?

PR #276, opened 2018-10-11
Catch all parse exceptions when parsing with scalameta

andrea-guarino-sonarsource:


andrea-guarino-sonarsource:
@pynicolas Added in f74571d73c06fe03959e385ed72b0aea49cfc138

PR #278, opened 2018-10-12
SONARSLANG-239 FP S107(TooManyParametersCheck) on overridden methods

andrea-guarino-sonarsource:


slang-checks/src/main/java/org/sonarsource/slang/checks/TooManyParametersCheck.java
```diff
@@ -62,4 +63,9 @@ public void initialize(InitContext init) {
     });
   }
 
+  private static boolean isOverrideMethod(FunctionDeclarationTree tree) {
+    return tree.modifiers().stream().anyMatch(mod ->
+      (mod instanceof ModifierTree) && ((ModifierTree) mod).kind() == ModifierTree.Kind.OVERRIDE);
```
pynicolas: These conditions are always true in the current test cases. We should add more test cases.

PR #281, opened 2018-10-12
Update rule metadata

alban-auzeill:


sonar-kotlin-plugin/src/main/resources/org/sonar/l10n/kotlin/rules/kotlin/Sonar_way_profile.json
```diff
@@ -16,7 +16,6 @@
     "S1145",
     "S1172",
     "S1186",
-    "S1186",
```
pynicolas: Why? That doesn't seem consistent with RSPEC.

PR #286, opened 2018-10-15
[Ruby] Fix "jar" filesystem created twice and not closed

alban-auzeill:


alban-auzeill:
deleted in favor of #287

PR #292, opened 2018-10-22
Move Ruby/Kotlin/Scala check lists from slang-checks to their respective plugins

alban-auzeill:


quentin-jaquier-sonarsource:
LGTM

PR #299, opened 2018-11-07
SONARSLANG-332 [Slang] Enhance TreeMetaDataProvider API to provide access to all tokens, index of tokens and update of token types

alban-auzeill:


slang-api/src/main/java/org/sonarsource/slang/impl/TreeMetaDataProvider.java
```diff
@@ -50,6 +51,44 @@ public TreeMetaDataProvider(List<Comment> comments, List<Token> tokens) {
...
+  public Optional<Token> previousToken(TextRange textRange) {
+    int textRangeIndex = indexOfFirstElement(sortedTokens, textRange);
+    if (textRangeIndex <= 0) {
+      return Optional.empty();
+    } else {
+      return Optional.of(sortedTokens.get(textRangeIndex - 1));
+    }
+  }
+
+  public void updateTokenType(Token token, Token.Type newType) {
```
pynicolas: It's a pity to break immutability. I hope that some other solutions were considered.

PR #305, opened 2018-11-12
SONARSLANG-341 Fix rule S1144 UnusedPrivateMethodCheck performance

christophe-zurn-sonarsource:


slang-checks/src/main/java/org/sonarsource/slang/checks/UnusedPrivateMethodCheck.java
```diff
@@ -52,27 +48,27 @@
...
+          usedIdentifiers.add(((IdentifierTree) tree));
+        }
+      });
+
+      usedIdentifiers.removeAll(methods.stream()
         .map(FunctionDeclarationTree::name)
         .collect(Collectors.toSet()));
 
-      classMethods.stream()
+      methods.stream()
```
alban-auzeill: Here we loop methods.size() x usedIdentifierNames.size() times.
I would prefer to collect only IdentifierTree.identifier() string in a set, and use "contains" method.

slang-checks/src/main/java/org/sonarsource/slang/checks/UnusedPrivateMethodCheck.java
```diff
@@ -52,27 +48,27 @@
...
         .filter(method -> FunctionUtils.isPrivateMethod(method) && !FunctionUtils.isOverrideMethod(method))
```
alban-auzeill: this filter can be moved above, when we add to "methods".

PR #307, opened 2018-11-13
SONARSLANG-290 Improve S1313: support port numbers IP addresses and IPv6

christophe-zurn-sonarsource:


christophe-zurn-sonarsource:
Note that files: `dependencies_test.rb` and `CipherSuitesTest.scala` are raising FP.
Although the strings raising issues are actually valid ipv6 addresses, it is not used as such in the context of these files.
However, given that this is a Hotspot rule, I believe it is OK as it is. WDYT?

PR #316, opened 2020-05-14
WIP: Include surefire-java parser into KotlinSensor

soudmaijer:
I want to get some feedback on my initial idea of adding the existing surefire parser to the KotlinSensor.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,6 +32,7 @@ allprojects {
         sonarqubeVersion = '7.3'
         orchestratorVersion = '3.26.0.2111'
         sonarlintVersion = '4.2.0.2266'
+        sonarJavaVersion = '6.3.0.21585'
         sonarLinksCi = 'https://travis-ci.org/SonarSource/slang'
         sonarLinksScm = 'https://github.com/SonarSource/slang'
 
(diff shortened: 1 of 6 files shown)


grumpytechdude:
Does this work? 

I'm very excited for this feature to be merged in, but I've no experience with SonarSource I'm afraid :(

margarita-nedzelska-sonarsource:
@soudmaijer thanks a lot for the contribution and your interest in our products.

On the one hand, your idea does make sense. We shouldn't write things from scratch if we already have existing implementation. So from this point of view reusing java surefire parser is smart.

On the other hand, as you've probably noticed, the dependency to java-surefire leads to the dependency on java-frontend which is something we want to avoid. Furthermore it makes kotlin-plugin twice bigger (we have size checks on purpose to make developers think about all the dependencies they are bringing to the plugin)

Another tricky part is not parsing itself, but report-to-sourcefile mapping. So this way java parser won't help, cause it visits and indexes java files only. And the problem is if sources are written in different languages.

Still this doesn't mean that this feature won't be implemented at all. Currently, I'am working on another solution to support surefire test reports for kotlin in sonarqube. And it will hopefully be in the next slang release.

Again, Thanks a lot for your PR. It helps us understand, what's important for our users. And we highly appreciate your contribution.

soudmaijer:
@margarita-nedzelska-sonarsource Thanks for picking this up, great to see sonarsource implemented this feature!

nuhkoca:
@margarita-nedzelska-sonarsource can you please give an update on this?

